PR #6, opened 2018-11-01
Add Synchronization primitive for synchronized access

vlovgr:
Pull request #5 introduced an `MVar` for having exclusive access to the `Consumer`, which is necessary for being able to close the `Consumer` safely (avoiding `ConcurrentModificationException`) when using an `ExecutionContext` with more than one thread in `ConsumerSettings`. 

This pull request replaces `MVar` with a `Synchronization` primitive based on `Ref` and `Deferred`.

Massive thanks to @SystemFw for the help leading up to this.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,6 +24,7 @@ lazy val docs = project
 
 lazy val dependencySettings = Seq(
   libraryDependencies ++= Seq(
+    "org.typelevel" %% "cats-effect" % "1.1.0-M1",
     "co.fs2" %% "fs2-core" % "1.0.0",
     "org.apache.kafka" % "kafka-clients" % "2.0.0"
   ),
--- a/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -17,7 +17,7 @@
 package fs2.kafka
 
 import cats.data.NonEmptyList
-import cats.effect.concurrent.{Deferred, MVar, Ref}
+import cats.effect.concurrent.{Deferred, Ref}
 import cats.effect.syntax.concurrent._
 import cats.effect.{ConcurrentEffect, Fiber, Timer, _}
 import cats.instances.list._
@@ -33,6 +33,7 @@ import cats.syntax.traverse._
 import fs2.concurrent.Queue
 import fs2.kafka.KafkaConsumerActor.Request._
 import fs2.kafka.KafkaConsumerActor._
+import fs2.kafka.internal.Synchronized
 import fs2.kafka.internal.syntax._
 import fs2.{Chunk, Stream}
 import org.apache.kafka.clients.consumer.{KafkaConsumer => KConsumer, _}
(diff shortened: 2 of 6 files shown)


SystemFw:
Ah, should have realised it was for this! :)

vlovgr:
@SystemFw I had to do one more change to instead bracket over `ref.getAndSet(next)`, so that `next` is always completed, even when `use` is cancelled. I've added a test for it as well. But we need cats-effect 1.1.0-M1 for the idempotency fix of bracket release.

SystemFw:
Yeah I was looking at that right now, and you're right, there are a couple of `flatMap`s that could be cancelled there

SystemFw:
One thing I haven't thought of (yet) though, is the efficiency of this solution

SystemFw:
One nice thing is that we avoid this issue 
https://github.com/typelevel/cats-effect/issues/382#issuecomment-433739248

vlovgr:
Yeah, that's quite nice. In terms of efficiency, it doesn't matter much here. But it is interesting, and I put together a very quick benchmark against `MVar`, just to check that it's not way out of line, and they at least look comparable.

```
Benchmark                             Mode  Cnt       Score      Error  Units
SynchronizationBench.measure1kMVar   thrpt   25  115719.477 ± 1855.768  ops/s
SynchronizationBench.measure1kSynch  thrpt   25  116473.836 ± 1069.626  ops/s
```

Code for the benchmark as follows.

```scala
package fs2.kafka.internal

import cats.effect.concurrent.MVar
import cats.effect.{Concurrent, ContextShift, Fiber, IO}
import cats.implicits._
import org.openjdk.jmh.annotations._

import scala.concurrent.ExecutionContext

class SynchronizationBench {
  private def useSynch(synch: Synchronized[IO, Unit])(
    implicit F: Concurrent[IO],
    context: ContextShift[IO]
  ): Int => IO[Fiber[IO, Unit]] =
    (n: Int) =>
      IO.shift >> synch
        .use { _ =>
          if (n % 100 == 0)
            IO.raiseError(new RuntimeException)
          else IO.unit
        }
        .attempt
        .void
        .start

  private def useMVar(mVar: MVar[IO, Unit])(
    implicit F: Concurrent[IO],
    context: ContextShift[IO]
  ): Int => IO[Fiber[IO, Unit]] =
    (n: Int) =>
      IO.shift >>
        F.bracket(mVar.take) { _ =>
            if (n % 100 == 0)
              IO.raiseError(new RuntimeException)
            else IO.unit
          }(mVar.put)
          .attempt
          .void
          .start

  private def testSynch(n: Int): Unit = {
    implicit val contextShift: ContextShift[IO] =
      IO.contextShift(ExecutionContext.global)

    Synchronized[IO]
      .of(())
      .flatMap { synchronized =>
        val use = useSynch(synchronized)
        (1 to n).toList.map(use).combineAll.flatMap(_.join)
      }
      .unsafeRunSync
  }

  private def testMVar(n: Int): Unit = {
    implicit val contextShift: ContextShift[IO] =
      IO.contextShift(ExecutionContext.global)

    MVar[IO]
      .of(())
      .flatMap { mVar =>
        val use = useMVar(mVar)
        (1 to n).toList.map(use).combineAll.flatMap(_.join)
      }
      .unsafeRunSync
  }

  @Benchmark
  @OperationsPerInvocation(1000)
  def measure1kSynch(): Unit = testSynch(1000)

  @Benchmark
  @OperationsPerInvocation(1000)
  def measure1kMVar(): Unit = testMVar(1000)
}
```

Benchmark ran with following arguments.

```
jmh:run -prof jmh.extras.JFR -t1 -f 1 -wi 10 -i 20 .*
```

SystemFw:
That's cool :)

PR #9, opened 2018-11-05
Add support for offset commit recovery

vlovgr:
Add `CommitRecovery` for defining offset commit recovery strategies. The default strategy only retries `RetriableCommitFailedException` with a jittered exponential backoff for 10 attempts, switching to fixed rate retries for up to another 5 attempts (see `CommitRecovery#Default`).

You can override the `CommitRecovery` using `ConsumerSettings#withCommitRecovery`. If you would like to go back to default behaviour before this change (no retries), simply use `withCommitRecovery` with `CommitRecovery#None`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,7 +63,18 @@ lazy val mimaSettings = Seq(
   mimaBinaryIssueFilters ++= {
     import com.typesafe.tools.mima.core._
     Seq(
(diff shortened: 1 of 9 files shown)


mwz:
This is very nice, 👍 for adding jitter.

PR #25, opened 2018-11-22
Update kafka-clients to 2.1.0

vlovgr:
Needs https://github.com/manub/scalatest-embedded-kafka/pull/160


vlovgr:
Going to split this into two pull requests, while we wait on 2.1.0.

PR #64, opened 2018-12-12
Expose `seek` function in KafkaConsumer

danielkarch:
Hi, 

for our application we need to be able to set the offset from which to start reading. I found that the `seek` function exists in the Java client but is not exposed in this library.

### DIFF
--- a/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -118,6 +118,19 @@ sealed abstract class KafkaConsumer[F[_], K, V] {
   def partitionedStream: Stream[F, Stream[F, CommittableMessage[F, K, V]]]
 
   /**
+    * Overrides the fetch offsets that the consumer will use on the next poll.
+    * If this API is invoked for the same partition more than once, the latest
+    * offset will be used on the next poll(). Note that you may lose data if
+    * this API is arbitrarily used in the middle of consumption, to reset the
+    * fetch offsets.
+    *
+    * The stream will raise an [[IllegalArgumentException]] if the provided
+    * offset is negative, or an [[IllegalStateException]] if the provided
+    * TopicPartition is not assigned to this consumer.
+    */
+  def seek(topicPartition: TopicPartition, offset: Long): Stream[F, Unit]
+
+  /**
     * Subscribes the consumer to the specified topics. Note that you have to
     * use one of the `subscribe` functions to subscribe to one or more topics
     * before using any of the provided `Stream`s, or a [[NotSubscribedException]]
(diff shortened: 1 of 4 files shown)


src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -155,5 +175,49 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
...
+      createCustomTopic(topic)
+
+      val produced = (0l until numMessages).map(n => s"key-$n" -> s"value->$n")
+      publishToKafka(topic, produced)
+
+      val consumed =
+          consumerSettings(config).flatMap { consumerSettings =>
+            consumerStream[IO].using(consumerSettings).flatMap { consumer =>
+
+              val validSeekParams =
```
danielkarch: This test setup is a bit ugly. Since `KafkaConsumer` does not expose e.g. `partitionsFor`, the only way I can get hold of a valid `TopicPartition`/`offset` pair is to commit a message and look inside the `CommitableOffset`.

danielkarch:
I addressed your comments.

src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -117,6 +117,19 @@ sealed abstract class KafkaConsumer[F[_], K, V] {
...
   def partitionedStream: Stream[F, Stream[F, CommittableMessage[F, K, V]]]
 
+  /**
+    * Overrides the fetch offsets that the consumer will use when reading the
+    * next message. If this API is invoked for the same partition more than once,
+    * the latest offset will be used. Note that you may lose data if this API is
+    * arbitrarily used in the middle of consumption to reset the fetch offsets.
+    *
+    * The stream obtained from [[stream]] or [[partitionedStream]] will raise
+    * an [[IllegalArgumentException]] if the provided offset is negative, or an
```
vlovgr: Scaladoc generation is failing because it can't find links for these exceptions. Easiest is to change to backticks (e.g. `IllegalArgumentException`) for both of these.

src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
vlovgr: We could always add `partitionsFor`. But I'm also happy to leave it as is for now and rewrite later.

vlovgr:
I'm thinking we should implement seek like we have done `beginningOffsets`, for example. We shouldn't have to wait until the stream to see any errors, and we shouldn't fail the consumer if seek fails (perhaps you want to recover from such errors and still be able to consume).

vlovgr:
I pushed a last couple of changes to get this over the line.

Thanks a lot for adding this! 👍

PR #79, opened 2019-01-23
Add methods to commit chunks with passthrough values

david-mcgillicuddy-ovo:
Currently missing any tests, is there interest in adding this? No idea there's anybody other than myself with this use case.

### DIFF
--- a/src/main/scala/fs2/kafka/CommittableOffsetBatch.scala
+++ b/src/main/scala/fs2/kafka/CommittableOffsetBatch.scala
@@ -117,8 +117,28 @@ object CommittableOffsetBatch {
     * per topic-partition.
     *
     * @see [[CommittableOffsetBatch#fromFoldableOption]]
+    * @see [[CommittableOffsetBatch#fromFoldableMap]]
     */
   def fromFoldable[F[_], G[_]](offsets: G[CommittableOffset[F]])(
+    implicit F: Applicative[F],
+    G: Foldable[G]
+  ): CommittableOffsetBatch[F] = fromFoldableMap[F, G, CommittableOffset[F]](offsets)(identity)
+
+  /**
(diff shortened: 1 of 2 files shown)


vlovgr:
@david-mcgillicuddy-ovo Thanks for raising this pull request. The implementation looks reasonable, and could be a good fit. I'm mostly curious about the use case? I haven't needed this myself yet, as we most often require at-least-once delivery guarantees, in which case the commit step is the last thing to happen (which is also why the current commit pipes return `Unit`).

david-mcgillicuddy-ovo:
That's a very good point, after some discussion it seems like we're probably just using it incorrectly.

One thing I do like about it is that you can have the strong guarantees internally (we definitely should be using Producer.producePassthrough) but then a secondary pipeline which logs metrics etc but that won't block the first or affect its commiting.

vlovgr:
@david-mcgillicuddy-ovo We typically include metrics etc. in the at-least-once guarantee, otherwise you run the risk of processing and committing records but never seeing them in your metrics. I'm not sure what you're using to record metrics, but with e.g. Kamon, metrics are buffered locally and only sent periodically in the background, so there's no blocking involved.

If you don't care about at-least-once semantics for things like metrics, you can always use fire-and-forget semantics with `start` (if such calls are blocking, then also shift to a dedicated context). Note that you can have a `F[CommittableOffset[F]]` as passthrough, where the `F` can e.g. record metrics.

```scala
// stream: Stream[F, ProducerRecord[F, String, String, F[CommittableOffset[F]]]]

stream
  .evalMap(producer.produce) // Stream[F, F[F[CommittableOffset[F]]]]
  .map(_.flatten) // Stream[F, F[CommittableOffset[F]]] -- produce and then record metrics, as one effect
  .through(commitBatchWithinF(500, 15.seconds)) // Stream[F, Unit]
```

PR #83, opened 2019-02-11
Add ConsumerSettings#withMaxPrefetchBatches

### LINKED ISSUE #82: Better document the backpressure and `pollInterval`
The documentation says that the (consumer) stream is backpressured. But after a quick look at the code base I cannot easily find such evidence. This makes it hard to set the `pollInterval` setting.

As far as I understand there are 3 pieces at work in the consumer stream:
1. the poll loop -- responsible for throttling the poll requests (the poll queue is bounded with 1 element)
2. the consumer loop (`KafkaConsumerActor`) -- handles both fetch requests issued by the stream as well as poll requests by the poll loop, using a `State` variable that accumulates fetch requests along with fetched kafka records.
3. the stream -- emits fetch requests on-demand (pull based)

At first it seems that the consumer loop simply accumulates endlessly records in the `State.records`. But then I came across the poll function in the consumer loop:
```scala
val resume = (requested intersect assigned) diff available
val pause = assigned diff resume

consumer.pause(pause.asJava)
consumer.resume(resume.asJava)
```
This seems like unless a fetch request is present no polling will be performed, in effect providing backpressure?

Perhaps some high-level documentation of the code-base would be useful?

My understanding is that `pollInterval` is only there to reduce the pressure on Kafka, if the record consumption rate (the logic that consumes the stream) exceeds the production rate (the kafka producer, up-stream)?

vlovgr:
This is an attempt to better document the backpressure in consumer streams. We add a new setting `maxPrefetchBatches` and corresponding setter `withMaxPrefetchBatches` to `ConsumerSettings`. An explanation of the backpressure and how this setting controls it is provided in the scaladoc.

Fixes #82.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,7 +45,7 @@ lazy val dependencySettings = Seq(
 lazy val mdocSettings = Seq(
   mdoc := run.in(Compile).evaluated,
   scalacOptions -= "-Xfatal-warnings",
-  crossScalaVersions := Seq(scalaVersion.value),
+  crossScalaVersions := Seq(scalaVersion.value)
 )
 
 lazy val buildInfoSettings = Seq(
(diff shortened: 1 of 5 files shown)


backuitist:
Hey thanks for this! Although this improves the situation I still thinks there's room for a README. The flow isn't trivial and I believe that despite the fact that READMEs tend to rot faster than scaladoc they do a better job at depicting the big picture.

vlovgr:
@backuitist Thanks for #84! I believe this should be added to the site instead of as a readme, but let's continue the conversation in #84.

PR #84, opened 2019-02-12
Add a README explaining the backpressure

backuitist:
It's not complete but perhaps a good starting point :)


backuitist:
I thought that the site would contain user focused documentation, whereas the README would contain instructions for contributors. I like when you don't have to click around to find the information :)
Feel free to move this to the documentation if you disagree though :)

vlovgr:
Thanks a lot for this, @backuitist! I believe it's nice to have all high-level documenation in one place. I'll take a stab at moving this to the site and adding some notes from myself as well, when I get some time.

backuitist:
Neat! Thanks!

PR #88, opened 2019-02-21
Add 'describeCluster' and 'createTopic' operations to KafkaAdminClient.

danxmoran:
I'm hoping to use `fs2-kafka` in a new project at work, but I need these two operations. I tried to follow the existing patterns in the `KafkaAdminClient`, but please let me know if I should change anything!

### DIFF
--- a/src/main/scala/fs2/kafka/KafkaAdminClient.scala
+++ b/src/main/scala/fs2/kafka/KafkaAdminClient.scala
@@ -24,7 +24,7 @@ import fs2.kafka.KafkaAdminClient._
 import fs2.kafka.internal.syntax._
 import org.apache.kafka.clients.admin._
 import org.apache.kafka.clients.consumer.OffsetAndMetadata
-import org.apache.kafka.common.{KafkaFuture, TopicPartition}
+import org.apache.kafka.common.{KafkaFuture, Node, TopicPartition}
 
 /**
   * [[KafkaAdminClient]] represents an admin client for Kafka, which is able to
@@ -34,6 +34,21 @@ import org.apache.kafka.common.{KafkaFuture, TopicPartition}
   * Use [[adminClientResource]] or [[adminClientStream]] to create an instance.
   */
 sealed abstract class KafkaAdminClient[F[_]] {
+
+  /**
+    * Create a new topic.
+    *
+    * NOTE: Kafka's base [[AdminClient]] supports creating multiple topics in a
+    * single request, but the result isn't transactional. We limit to a single
+    * topic-per-request to avoid complications in error reporting for cases when
+    * some topic creations fail and others succeed.
+    */
+  def createTopic(topic: NewTopic): CreateTopic[F]
+
+  /**
+    * Describes nodes in the current cluster.
+    */
+  def describeCluster: DescribeCluster[F]
 
   /**
     * Describes the consumer groups with the specified group ids, returning a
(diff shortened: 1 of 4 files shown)


src/main/scala/fs2/kafka/KafkaAdminClient.scala
```diff
@@ -35,6 +35,21 @@ import org.apache.kafka.common.{KafkaFuture, TopicPartition}
   */
 sealed abstract class KafkaAdminClient[F[_]] {
 
+  /**
+    * Create a new topic.
+    *
+    * NOTE: Kafka's base [[AdminClient]] supports creating multiple topics in a
+    * single request, but the result isn't transactional. We limit to a single
+    * topic-per-request to avoid complications in error reporting for cases when
+    * some topic creations fail and others succeed.
```
danxmoran: How do you all feel about this design choice? I couldn't think of an obvious way to convert the `Map[String, Future[Void]]` of topic-creation results provided by Kafka into a single `F[_]` capturing all failures and all successes, and figured it's easy enough for users to build that behavior however they'd like using Cats and fs2.

src/main/scala/fs2/kafka/KafkaAdminClient.scala
```diff
@@ -113,6 +128,32 @@ sealed abstract class KafkaAdminClient[F[_]] {
 }
 
 object KafkaAdminClient {
+
+  sealed abstract class CreateTopic[F[_]] {
+
+    /** Name of the topic being created. */
+    def topicName: String
```
danxmoran: I added this to make error reporting easier if you've got a collection of `CreateTopic[F]`s

src/main/scala/fs2/kafka/KafkaAdminClient.scala
vlovgr: We can do something similar to what we do for e.g. `describeTopics`:

```scala
def createTopics[G[_]](topics: G[NewTopic])(
  implicit G: Foldable[G]
): F[Unit]
```

and `createTopicsWith` then simply becomes the following.

```scala
private[this] def createTopicsWith[F[_], G[_]](
  client: Client[F],
  topics: G[NewTopic]
)(implicit G: Foldable[G]): F[Unit] =
  client(_.createTopics(topics.asJava).all.void)
```

You might need to add `void` to `KafkaFutureSyntax` in `fs2.kafka.internal.syntax`.

```scala
def void: KafkaFuture[Unit] =
  map(_ => ())
```

danxmoran: :+1: to using `Foldable` to handle the input collection, but my worry is about the `.all` call. The underlying `KafkaFuture.allOf` arbitrarily chooses a single error to bubble up if multiple Futures have failed. For our use-case, we'd need to be able to collect & report every failure.

If there's interest on your end, I can spend some time setting up more complex error-handling (i.e. using a custom `Exception` type that can collect all the errors). Otherwise I'll make the change as you suggest, and we'll just only call the method with one topic at a time.

vlovgr: I think it's difficult to cater for all use cases here, but it's pretty easy to write yourself using only `createTopics` (or `createTopic` for a single topic), `parTraverse`, and `attempt`. You can also easily include the topic name, or other details you care about.

```scala
topics.parTraverse(topic => client.createTopic(topic).attempt)
// e.g. IO[List[Either[Throwable,Unit]]]
```

I've pushed some changes, including `createTopic`. Would you mind having a look?

src/main/scala/fs2/kafka/KafkaAdminClient.scala
```diff
@@ -113,6 +146,18 @@ sealed abstract class KafkaAdminClient[F[_]] {
 }
 
 object KafkaAdminClient {
+  private[this] def createTopicWith[F[_]](
+    client: Client[F],
+    topic: NewTopic
+  ): F[Unit] =
+    client(_.createTopics(java.util.Collections.singleton(topic)).all.void)
```
danxmoran: Would it make sense for this to instead call `createTopicsWith(client, List(topic))`, to avoid the duplicate logic?

vlovgr: We could, but we avoid creating the extra intermediate `List` by doing it this way.

danxmoran:
Thanks for the help! LGTM, should I push a few more tests to make codecov happy?

vlovgr:
Looks like it was me who missed the additional tests -- pushed now.

Thanks a lot for this contribution! 👍  I'll merge this and get a new release out today.

PR #90, opened 2019-02-22
WIP: Bump version of kafka-clients to 2.1.1

sebastianvoss:
Bump version of kafka-clients to 2.1.1 to get the fix for https://issues.apache.org/jira/browse/KAFKA-7890 (shop-stopper bug when using kafka-clients in environments where brokers can change their IPs, e.g. Kubernetes).

For some reasons test in `KafkaAdminClientSpec` are red. Maybe we need to wait for `embeddedkafka` to release a build for Kafka 2.1.1?!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import ReleaseTransformations._
 
 val fs2Version = "1.0.3"
 
-val kafkaVersion = "2.1.0"
+val kafkaVersion = "2.1.1"
 
 lazy val `fs2-kafka` = project
   .in(file("."))


vlovgr:
Thanks for this @sebastianvoss! In my experience, the embeddedkafka version has to match exactly with the Kafka client version for it to work. Looks like the update has already been merged ([#35](https://github.com/embeddedkafka/embedded-kafka/pull/35)), but there's no release yet.

As a temporary workaround, 2.1.1 should be binary-compatible with 2.1.0, so if you add it as an explicit dependency in your project, everything should hopefully still work.

vlovgr:
@sebastianvoss Looks like it works if we override the Kafka version in tests. I've branched and included this fix in #91.

PR #104, opened 2019-03-26
Update to Kafka 2.2.0

CremboC:
I've updated to Kafka 2.2.0.

There was a deprecation so I fixed that. Other than seems like there were no issues - tests passing etc.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import ReleaseTransformations._
 
 val fs2Version = "1.0.4"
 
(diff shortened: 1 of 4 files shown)


vlovgr:
Thanks for this @CremboC! We can use `FiniteDurationSyntax#asJava` for the conversions -- I've pushed a commit to change this.

CremboC:
Ah of course, sorry about that! Thanks for the change.

PR #106, opened 2019-03-28
Change to run rebalance callbacks during poll

### LINKED ISSUE #105: Race condition between `subscribe` and `stream` leading to duplicated records
When subscribing to a topic the `KafkaConsumerActor` adds a rebalancing hook that registers `FetchRequest`s.
Then when calling `stream` the `KafkaConsumer` will requests partition assignments and - that's the race condition - depending on the timing might get some partition ids for which it will also register `FetchRequest`s.
This is aggravated by the fact that the `KafkaConsumerActor.State` does not restrict the amount of `FetchRequest`s per (partition, consumer), leading to duplicated records sent downstream.

vlovgr:
Issue #105 describes a situation where the following happens.

1. Assignment of `x` happens, an `Assigned(x)` message is put on the Actor's queue.
2. An `Assignment` request is processed and returns `x` as current assignment.
    This also registers a rebalance listener for assignments.
3. Streams are enqueued for the current assignment `x`.
4. The rebalance listener enqueues streams for assignment `x`.

Steps 3 and 4 can happen in reverse order, and there can be more assignments happening at the same time, but this covers the issue. The reason this happens is due to the fact that rebalance listeners simply queue messages onto the Actor's queue, rather than running the rebalance callbacks. This pull request changes to synchronously run the rebalance callbacks in the rebalance listener.

This also fixes another issue: we currently break the `ConsumerRebalanceListener` guarantee that `onPartitionsRevoked` is called before `onPartitionsAssigned` by asyncronously enqueuing the `Assigned` and `Revoked` messages (`Assigned` can potentially happen before `Revoked`).

Fixes #105.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -155,7 +155,11 @@ lazy val mimaSettings = Seq(
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaConsumer.assignment"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaConsumer.position"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaConsumer.seekToBeginning"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaConsumer.seekToEnd")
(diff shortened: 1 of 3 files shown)


backuitist:
Looks good, I like that we get rid of a couple of actor messages :)

PR #107, opened 2019-03-28
Multiple fetch request

backuitist:


backuitist:
Should complement well https://github.com/ovotech/fs2-kafka/pull/106.

src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -358,7 +360,10 @@ private[kafka] object KafkaConsumer {
         actorFiber combine pollsFiber
       }
 
+      private val streamIdCounter = new AtomicInteger(0)
+
       override def partitionedStream: Stream[F, Stream[F, CommittableMessage[F, K, V]]] = {
+        val streamId = streamIdCounter.getAndIncrement()
```
vlovgr: Unfortunately, this breaks referential transparency. `Ref[F, Int]` and `modify` is the alternative.

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -491,8 +492,10 @@ private[kafka] object KafkaConsumerActor {
       deferred.complete((chunk, FetchCompletedReason.FetchedRecords))
   }
 
+  type StreamId = Int
```
vlovgr: There are a few places in the pull request where we don't use this alias, but just say `Int` instead.

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -502,10 +505,18 @@ private[kafka] object KafkaConsumerActor {
...
       deferred: Deferred[F, (Chunk[CommittableMessage[F, K, V]], FetchCompletedReason)]
     ): State[F, K, V] = {
-      val fetch = NonEmptyChain.one(FetchRequest(deferred))
-      copy(fetches = fetches combine Map(partition -> fetch))
+      val fetch = Map(streamId -> FetchRequest(deferred))
+
+      // revoke any previous fetch on the same (partition,streamId)
+      // TODO log a warning as this shouldn't happen!
+      val existing = fetches.get(partition).flatMap(_.get(streamId))
+      existing.foreach(_.completeRevoked(Chunk.empty))
```
vlovgr: `completeRevoked` returns `F[Unit]` which never runs here.

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -502,10 +505,18 @@ private[kafka] object KafkaConsumerActor {
...
+
+      val fetchesForPartition = fetches.getOrElse(partition, Map.empty) ++ fetch
```
vlovgr: We've already done the `Map` lookup as part of `existing`, so we could reuse it here. We could also use `updated` to avoid creating the extra `Map` for `fetch`.

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -502,10 +505,18 @@ private[kafka] object KafkaConsumerActor {
...
+      copy(fetches = fetches ++ Map(partition -> fetchesForPartition))
```
vlovgr: Same here, could use `updated` instead to avoid creating the extra `Map`.

src/main/scala/fs2/kafka/KafkaConsumer.scala
backuitist: How do you plan to use `Ref[F, Int]` to avoid breaking referential transparency while not changing the interface?

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
backuitist: This type alias is only here for documentation purpose:
```
fetches: Map[TopicPartition, Map[StreamId, FetchRequest[F, K, V]]],
```
instead of
```
fetches: Map[TopicPartition, Map[Int, FetchRequest[F, K, V]]],
```
Everywhere else, we have a named parameter: `streamId: StreamId`. I you'd want to be type-safe you'd create another type wrapping Int, but since this is an internal API I fine with this (saving some memory).

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
backuitist: Good catch! Will fix

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
backuitist: Sure

src/main/scala/fs2/kafka/KafkaConsumer.scala
vlovgr: We create the `Ref[F, Int]` as part of `consumerResource`, and pass it to `createKafkaConsumer`. In `partitionedStream` we then add something like the following.

```scala
Stream.eval(ref.modify(n => (n + 1, n))).flatMap { streamId =>
  // ...
}
```

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
vlovgr: I meant that we could say `StreamId` everywhere where we say `Int`, including e.g. `streamId: StreamId`, `Ref[F, StreamId]`, etc.

src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -246,15 +246,23 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
     val assigned =
       withConsumer { consumer =>
         F.delay(consumer.assignment.contains(partition))
       }
 
-    def storeFetch =
-      ref.update(_.withFetch(partition, deferred))
+    def storeFetch = {
+      // revoke any previous fetch on the same (partition,streamId)
+      // TODO log a warning as this shouldn't happen!
```
vlovgr: Initial logging support is now in `master` as part of #113.

backuitist: Good! I'll finish the PR tomorrow (I'm off today).

vlovgr: I found another case where we send duplicate requests (downstream finishes), so I pushed a fix for that, along with rebase on master. The tests now pass without any warning log from here. Would be great if you could have a look through the pull request tomorrow. 👍

backuitist:
Thanks for fixing my commits! Looks all good to me :)

PR #109, opened 2019-03-29
Change to use updated state after poll

vlovgr:
In #106 we changed to run rebalance listeners as part of poll, but forgot to change to use the updated state after poll, and we were still relying on the previous state.


backuitist:
Yeah that `state` variable got me puzzled. Good that it's gone :)

PR #116, opened 2019-04-03
Improve readability of some functions by use for-comprehensions instead of nested `flatMap`

CremboC:
This is a WIP until I do more of them. I'm trying a function-per-commit so it's easier to see changes.


vlovgr:
Thanks for this @CremboC!

PR #117, opened 2019-04-03
Change to handle commits during rebalancing

vlovgr:


ennru:
@vlovgr Do you have any results of how this solution behaves with real data? I'm still hesitant if https://github.com/akka/alpakka-kafka/pull/751 is ready do go.

vlovgr:
@ennru @CremboC have tested this while scaling up consumers, and rebalancing seem to work well, compared to before where `CommitFailedException` was so prevalent it was difficult to get a stable consumer group. @CremboC might be able to provide more details.

Also, huge thanks for the detailed write-up and pull-request @ennru!

CremboC:
@ennru this PR definitely fixed the issues that I saw.

Essentially, I couldn't use this library at all at our scale and with our configuration with more than a single consumer. Everything worked perfectly fine, but then when the service had to scale up to multiple consumers (i.e. number of messages per second became too high for a single consumer to consume), everything went to hell. They would constantly die with `CommitFailedException`, as @vlovgr said and would take anywhere between 5 minute and never (yes, never) to fully recover. Every new deployment was the same thing over again - even if there was a single consumer in the previous deployment.

With this PR merged, everything works perfectly - no issues during scale up at all.

The only thing not 100% fixed as far as I'm concerned is I'm still seeing quite a bit of `Node 5 was unable to process the fetch request with (sessionId=1545249616, epoch=INITIAL): INVALID_FETCH_SESSION_EPOCH` "errors", but I'm not sure if it's related or not.

If you need any numbers, let me know 👍

PR #126, opened 2019-05-03
Simplify callback code in KafkaProducer#produce

kubukoz:
Because there's only one place where the promise could be completed, I figured Deferred could be replaced with a simple `async`. Tests pass (locally), so I'm submitting, but if there's a specific reason why Deferred was used feel free to close this ;)

### DIFF
--- a/src/main/scala/fs2/kafka/KafkaProducer.scala
+++ b/src/main/scala/fs2/kafka/KafkaProducer.scala
@@ -67,10 +67,10 @@ private[kafka] object KafkaProducer {
         .create(settings)
     } { producer =>
       F.delay {
-        producer.close(settings.closeTimeout.asJava)
-      }
-      .start
-      .flatMap(_.join)
+          producer.close(settings.closeTimeout.asJava)
+        }
+        .start
+        .flatMap(_.join)
     }
   }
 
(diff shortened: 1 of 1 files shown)


kubukoz:
Worth noting: this would enable lowering the constraints in `KafkaProducer.producerResource` and friends to `F: Concurrent`, albeit that'd be very binary incompatible. If need be, overloads could be introduced (and the existing methods deprecated) or that could wait until the next binary-breaking release.

vlovgr:
Thanks for the pull request!

It would be great to only require `Concurrent`, and there have been [previous attempts](https://github.com/ovotech/fs2-kafka/issues/102). The reason we require more is that we decouple calling `producer.send` and 'waiting for the record to be sent', as per the documentation of `produce`.

> Produces the `ProducerRecord`s in the specified `ProducerMessage` in two steps: the first effect puts the records in the buffer of the producer, and the second effect waits for the records to have been sent.

In your case, the outer `F.delay` is redundant, and you could write a `produce` function which both calls `producer.send` and 'waits for the record to be sent' using only `Async`. We decouple the two for performance reasons. The producer works in batches, and so you want to call `producer.send` for multiple records, to achieve batching, before waiting for any records to be sent.

Seeing as this is not the first time this has come up, we should definitely have a test for it. :)

kubukoz:
Yeah, for some reason when I was submitting this I forgot `async` is suspended anyway... so yeah, a test would help :D I'll try and contribute it when I have a moment

PR #130, opened 2019-05-08
Add new transactional producer

### LINKED ISSUE #128: Support for transactions
Kafka has [special APIs](https://www.confluent.io/blog/transactions-apache-kafka/) to support transactional read-process-write message flows. The gist seems to be:
1. Producers get an extra piece of config `transactional.id`, and call `initTransactions()` on startup
2. Committing offsets runs through the producer, which surrounds each batch with begin-/commit-transaction calls

What would be the best way to add this functionality to `fs2-kafka`? I was looking through Alpakka for reference and it looks like they separate "normal" and transactional types for both producers and consumed messages. I like the idea of using types to force the separation of use-cases, but I'm not sure how much of a maintenance burden it'd create for this project.

My team has a need for this functionality, so I'd be happy to help implement the new features.

danxmoran:
Fixes #128.

Here's an implementation which (mostly) avoids touching existing classes. There's a bunch of code copy-pasted from the existing `KafkaProducer` and `ProducerMessage` which might be nice to consolidate 😄

### DIFF
--- a/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -16,14 +16,15 @@
 
 package fs2.kafka
 
-import cats.effect.Sync
 import cats.Show
+import cats.effect.Sync
 import org.apache.kafka.clients.consumer.{Consumer, ConsumerConfig}
-import org.apache.kafka.common.requests.OffsetFetchResponse
+import org.apache.kafka.common.requests.{IsolationLevel, OffsetFetchResponse}
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
+
 import scala.collection.JavaConverters._
-import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 
 /**
   * [[ConsumerSettings]] contain settings necessary to create a
@@ -222,6 +223,18 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
     * }}}
     */
   def withDefaultApiTimeout(defaultApiTimeout: FiniteDuration): ConsumerSettings[F, K, V]
+
+  /**
+    * Returns a new [[ConsumerSettings]] instance with the specified
+    * isolation level. This is equivalent to setting the following property
+    * using the [[withProperty]] function, except you can specify it
+    * with an `IsolationLevel` instead of a `String`.
+    *
+    * {{{
+    * ConsumerConfig.ISOLATION_LEVEL_CONFIG
+    * }}}
+    */
+  def withIsolationLevel(isolationLevel: IsolationLevel): ConsumerSettings[F, K, V]
 
   /**
     * Includes a property with the specified `key` and `value`.
@@ -462,6 +475,15 @@ object ConsumerSettings {
       withProperty(
         ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG,
         defaultApiTimeout.toMillis.toString
+      )
+
+    override def withIsolationLevel(isolationLevel: IsolationLevel): ConsumerSettings[F, K, V] =
+      withProperty(
+        ConsumerConfig.ISOLATION_LEVEL_CONFIG,
+        isolationLevel match {
+          case IsolationLevel.READ_COMMITTED   => "read_committed"
+          case IsolationLevel.READ_UNCOMMITTED => "read_uncommitted"
+        }
       )
 
     override def withProperty(key: String, value: String): ConsumerSettings[F, K, V] =
--- a/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -16,15 +16,14 @@
 
 package fs2.kafka
 
-import cats.Show
 import cats.effect.Sync
+import cats.Show
 import org.apache.kafka.clients.consumer.{Consumer, ConsumerConfig}
-import org.apache.kafka.common.requests.{IsolationLevel, OffsetFetchResponse}
+import org.apache.kafka.common.requests.OffsetFetchResponse
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
-
 import scala.collection.JavaConverters._
-import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext
 
 /**
   * [[ConsumerSettings]] contain settings necessary to create a
@@ -231,9 +230,9 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
 
   /**
     * Returns a new [[ConsumerSettings]] instance with the specified
-    * isolation level. This is equivalent to setting the following property
-    * using the [[withProperty]] function, except you can specify it
-    * with an `IsolationLevel` instead of a `String`.
+    * isolation level. This is equivalent to setting the following
+    * property using the [[withProperty]] function, except you can
+    * specify it with an [[IsolationLevel]] instead of a `String`.
     *
     * {{{
     * ConsumerConfig.ISOLATION_LEVEL_CONFIG
@@ -488,8 +487,8 @@ object ConsumerSettings {
       withProperty(
         ConsumerConfig.ISOLATION_LEVEL_CONFIG,
         isolationLevel match {
-          case IsolationLevel.READ_COMMITTED   => "read_committed"
-          case IsolationLevel.READ_UNCOMMITTED => "read_uncommitted"
+          case IsolationLevel.ReadCommittedIsolationLevel   => "read_committed"
+          case IsolationLevel.ReadUncommittedIsolationLevel => "read_uncommitted"
         }
       )
 
@@ -568,7 +567,7 @@ object ConsumerSettings {
             byteArrayDeserializer,
             byteArrayDeserializer
           )
-      }
+        }
     )
 
   /**
--- a/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -560,7 +560,7 @@ object ConsumerSettings {
             byteArrayDeserializer,
             byteArrayDeserializer
           )
-        }
+      }
     )
 
   /**
--- a/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -568,7 +568,7 @@ object ConsumerSettings {
             byteArrayDeserializer,
             byteArrayDeserializer
           )
-        }
+      }
     )
 
   /**
--- a/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -127,6 +127,11 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
   def withGroupId(groupId: String): ConsumerSettings[F, K, V]
 
   /**
+    * The group ID consumers created with these settings will report.
+    */
+  def groupId: Option[String]
+
+  /**
     * Returns a new [[ConsumerSettings]] instance with the specified
     * max poll records. This is equivalent to setting the following
     * property using the [[withProperty]] function, except you can
@@ -438,6 +443,8 @@ object ConsumerSettings {
     override def withGroupId(groupId: String): ConsumerSettings[F, K, V] =
       withProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId)
 
+    override def groupId: Option[String] = properties.get(ConsumerConfig.GROUP_ID_CONFIG)
+
     override def withMaxPollRecords(maxPollRecords: Int): ConsumerSettings[F, K, V] =
       withProperty(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxPollRecords.toString)
 
@@ -561,7 +568,7 @@ object ConsumerSettings {
             byteArrayDeserializer,
             byteArrayDeserializer
           )
-      }
+        }
     )
 
   /**
--- a/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -126,11 +126,6 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
   def withGroupId(groupId: String): ConsumerSettings[F, K, V]
 
   /**
-    * The group ID consumers created with these settings will report.
-    */
-  def groupId: Option[String]
-
-  /**
     * Returns a new [[ConsumerSettings]] instance with the specified
     * max poll records. This is equivalent to setting the following
     * property using the [[withProperty]] function, except you can
@@ -441,8 +436,6 @@ object ConsumerSettings {
 
     override def withGroupId(groupId: String): ConsumerSettings[F, K, V] =
       withProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId)
-
-    override def groupId: Option[String] = properties.get(ConsumerConfig.GROUP_ID_CONFIG)
 
     override def withMaxPollRecords(maxPollRecords: Int): ConsumerSettings[F, K, V] =
       withProperty(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxPollRecords.toString)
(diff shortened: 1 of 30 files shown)


src/main/scala/fs2/kafka/ConsumerSettings.scala
```diff
@@ -206,6 +209,16 @@ sealed abstract class ConsumerSettings[K, V] {
...
 
+  /**
+    * Returns a new [[ConsumerSettings]] instance with read isolation level
+    * set to either "read_committed" (`true`) or "read_uncommitted" (`false`).
+    *
+    * {{{
+    * ConsumerConfig.ISOLATION_LEVEL_CONFIG
+    * }}}
+    */
+  def withReadCommittedIsolation(onlyReadCommitted: Boolean): ConsumerSettings[K, V]
```
danxmoran: Wasn't sure if I should make this a `Boolean` or make a new type similar to `Acks`... thoughts?

src/main/scala/fs2/kafka/TransactionalMessage.scala
```diff
@@ -0,0 +1,171 @@
...
+  * The [[passthrough]] and [[records]] can be retrieved from an
+  * existing [[TransactionalMessage]] instance.<br>
+  * <br>
+  * For a [[TransactionalMessage]] to be usable by [[TransactionalKafkaProducer]],
+  * it needs a `Traverse[F]` instance. This requirement is
+  * captured in [[TransactionalMessage]] as [[traverse]].
+  */
+sealed abstract class TransactionalMessage[F[_], G[+ _], +K, +V, +P] {
+
+  def records: G[(ProducerRecord[K, V], CommittableOffset[F])]
```
danxmoran: Instead of this, I considered making `TransactionalMessage` a subclass of `ProducerMessage` adding a `def committableOffsetBatch: CommittableOffsetBatch[F]` field. Thoughts?

src/main/scala/fs2/kafka/package.scala
```diff
@@ -541,4 +541,33 @@ package object kafka {
...
+    transactionalId: String,
+    consumerGroupId: String,
+    blockingEc: ExecutionContext
+  )(
+    implicit F: ConcurrentEffect[F],
+    C: ContextShift[F]
+  ): Stream[F, TransactionalKafkaProducer[F, K, V]] =
+    Stream.resource(
+      transactionalProducerResource(settings, transactionalId, consumerGroupId, blockingEc)
+    )
```
danxmoran: I can add the other helpers once we're happy with the APIs

src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -0,0 +1,185 @@
...
+      .evalMap { producer =>
+        C.evalOn(blockingEc)(F.delay(producer.initTransactions())).as(producer)
+      }
+  }
+
+  def producerResource[F[_], K, V](
+    settings: ProducerSettings[K, V],
+    transactionalId: String,
+    consumerGroupId: String,
+    blockingEc: ExecutionContext
```
danxmoran: `transactionalId` and `blockingEc` could go into a new `TransactionalSettings` type. `consumerGroupId` could probably be a field on `TransactionalMessage` instead of a per-consumer setting. Thoughts?

src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -0,0 +1,185 @@
...
+  )(
+    implicit F: ConcurrentEffect[F],
+    C: ContextShift[F]
+  ): Resource[F, TransactionalKafkaProducer[F, K, V]] =
+    createProducer(settings, transactionalId, blockingEc).map { producer =>
+      new TransactionalKafkaProducer[F, K, V] {
+        override def produce[G[+ _], P](
+          message: TransactionalMessage[F, G, K, V, P]
+        ): F[F[ProducerResult[G, K, V, P]]] =
```
danxmoran: The double-`F` wrapper isn't as useful here because (I think) committing a transaction forces a buffer flush within the producer. Is it worth keeping, or should we just flatten it out for clarity?

src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
vlovgr: I think we can re-use `ProducerSettings`, since we should get an error on `initTransactions` if we forget to set `transactionalId`. So let's add `withTransactionalId` to `ProducerSettings`. In the `0.20.x` branch, we already have a `ExecutionContext` on `ProducerSettings`, so maybe it's worth rebasing against that branch?

Regarding `consumerGroupId`, I'm thinking maybe we should put it on `CommittableOffset` (and `CommittableOffsetBatch`), so the user only has to specify it once for the consumer and nowhere else.

src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
vlovgr: Agreed, we should just return a single effect here.

src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
danxmoran: I think tracking the group ID in `CommittableOffset` will definitely work. For `CommittableOffsetBatch`, I'm worried there isn't a good default behavior for:
* Creating a batch from two offsets with different group IDs
* Merging two batches with different group IDs

In the transactional case these mismatches should raise errors, but for non-transactional setups they seem harmless.

My thoughts for a middle ground are:
* Track `consumerGroupId` within `CommittableOffset`
* Rework the builder methods for `TransactionalMessage` to validate that all the group IDs match (and return `F[TransactionalMessage]`)

vlovgr: Sounds like a good middle ground. 👍

danxmoran:
@vlovgr I think I covered everything

src/main/scala/fs2/kafka/TransactionalProducerMessage.scala
```diff
@@ -0,0 +1,208 @@
...
+    record: (ProducerRecord[K, V], CommittableOffset[F])
+  )(implicit F: MonadError[F, Throwable]): F[TransactionalProducerMessage[F, Id, K, V, Unit]] =
+    one(record, ())
+
+  /**
+    * Creates a new [[TransactionalProducerMessage]] for producing zero or more
+    * `ProducerRecord`s within a Kafka transaction, then emitting a
+    * [[ProducerResult]] with the results and specified passthrough value.
+    */
+  def apply[F[_], G[+ _], K, V, P](
```
vlovgr: With this approach, we have to have exactly one `ProducerRecord` for a `CommittableOffset`. I think there is a case that you might actually have zero or more `ProducerRecord`s for every `CommittableOffset`, but currently we can't support that. Do you have any thoughts on this and how we could support it?

danxmoran: Ergh, good point... I knew this would drop the zero-`ProducerRecord`s case (which seems OK since transactions are designed for consume-process-produce flows), but hadn't considered the many-`ProducerRecord`s case.

I could see two approaches to a different API:
1. Change this to take both `records: G[ProducerRecord[K, V]]` and `offsets: G[CommittableOffset[F]]`. This lets us avoid changing anything in `CommittableOffsetBatch`, but it forces the same `G` context for both args (which might not always make sense?)
2. Change this to take `records: G[ProducerRecord[K, V]]` and `offsetBatch: CommittableOffsetBatch[F]`, and add tracking of group IDs to `CommittableOffsetBatch` like you previously suggested. To get around the concerns I raised before I think the batch would need to track something like `consumerGroupIds: Set[String]` instead of just `consumerGroupId: Option[String]`. `TransactionalProducerMessage.apply` could enforce that the set contains exactly one element, but other usage sites could be OK with group ID sets of zero / many elements.

I lean towards trying option 2. WDYT?

vlovgr: I gave this some thought, and if we look at it from a user point-of-view, you would use it something along the following lines. I'm assuming we want to commit multiple offsets per transaction, so you might use `Stream#chunks` or similar for chunking.

```scala
messages: Chunk[CommittableMessage[F, K, V]] =>
  TransactionalProducerMessage(
    records = messages.flatMap(message => Chunk(ProducerRecord(...))), // Technically we need NonEmptyTraverse
    offsetBatch = CommittableOffsetBatch.fromFoldable(messages.map(_.committableOffset)),
    passthrough = ???
  ).flatMap(producer.produce)
```

Two things which I think we could improve on here are the following.

1. We `map` over `messages` twice and there's a disconnect between which offsets are for which `ProducerRecord`s. It doesn't matter for the technical implementation, but for a user, I think it's nice to only have to do one `map` and see which offset is for which `ProducerRecord`s.
2. Due to the error handling being on `TransactionalProducerMessage`, we have to `flatMap`. This is not a big annoyance, but it results in less straightforward code than if error handling was done in `producer.produce`, since you wouldn't need to `flatMap` in that case.

If we wrote the equivalent code but taking those two points into account, we would end up with something like the following.

```scala
messages: Chunk[CommittableMessage[F, K, V]] =>
  producer.produce(
    TransactionalProducerMessage(
      messages.map { message =>
        CommittableProducerRecords(
          Chunk(ProducerRecord(...)),
          message.committableOffset
        )
      },
      passthrough = ???
    )
  )
```

The introduction of `CommittableProducerRecords` means we can see which offset is for which `ProducerRecord`s, and there's now only one `map`. Error handling is in `producer.produce` and I think we can handle the empty case with `F.unit` (since there's nothing to do).

`TransactionalProducerMessage` can do the offset batching using the exsiting `Traverse[G]` instance, but the slightly more tricky part is that we have `G[CommittableProducerRecords[F, H, K, V]]`, from which we can get `G[H[ProducerRecord[K, V]]`. This has to be 'flattened' to `G[ProducerRecord[K, V]]`. I think we can require `H = G`, so we can flatten easily. If type inference works, we can even have a `CommittableProducerRecords.one` which does the wrapping with `G` using `Applicative[G]`.

The added benefit is that some offsets in our transaction can now not result in producing some records, which might not be common, but is nice to support.

What do you think? I'm also happy to take a stab at this myself unless you want to.

danxmoran: I think trying to reduce the number of `map` calls needed makes sense :+1: `CommittableProducerRecords` might also "fix" the issue of handling messages from multiple consumer groups in one transaction, if we have code within the `TransactionalKafkaProducer` like:
```scala
message: TransactionalProducerMessage =>
  message.committableRecords.traverse { records =>
    underlyingProducer.sendOffsetsToTransaction(
      records.offset.offsets,
      records.offset.consumerGroupId
    )
  }
```
... But that loses the efficiency of batching the offsets, so we might still want that as an internal pre-processing step.

> we can handle the empty case with F.unit (since there's nothing to do).

I think in this case we'd still need to call `commit` on the underlying offsets. Otherwise users will end up needing to branch the empty & non-empty cases anyways to ensure all the input records get committed.

I won't have much time the next few days to work on this PR, but I'm happy to try it out later if you don't get to it first 😄

vlovgr: > ... But that loses the efficiency of batching the offsets, so we might still want that as an internal pre-processing step.

It's a good point. It should be rare to mix records from multiple consumers (`CommittableOffsetBatch` cannot even handle it properly at the moment). It's a good idea to guard against it though, and hopefully we can have `CommittableOffsetBatch` guard against it in the future as well.

> I think in this case we'd still need to call `commit` on the underlying offsets. Otherwise users will end up needing to branch the empty & non-empty cases anyways to ensure all the input records get committed.

We have two cases: (1) the `Chunk` is empty, so you end up with a `TransactionalProducerMessage` without any `CommittableProducerRecords` and only a passthrough, in which case you can do `F.pure(message.passthrough)` (this is the case I was referring to), (2) the `Chunk` is non-empty, but there are `CommittableProducerRecords` without records in them, in which case we have to `commit` (the case you're probably referring to).

> I won't have much time the next few days to work on this PR, but I'm happy to try it out later if you don't get to it first 😄

I'm quite interested to see if this works well, will try to find some time for it.

vlovgr: I've pushed some WIP which passes the tests. Still need more tests, documentation, and general cleanup, but it gives an idea of the approach.

danxmoran: I think this looks good, thanks for all the help! I started working on more tests, and I'm finding that the choice to require `G[_]` to be the same in both `CommittableRecords` and `TransactionalProducerMessage` can be a rough spot in usability. For example, the code I'd like to write is:
```scala
toProduce: List[(K, V, CommittableOffset[F])] =>
  val records = toProduce.map {
    case (key, value, offset) =>
      CommittableProducerRecords.one(ProducerRecord(topic, key, value), offset)
  }
  val message = TransactionalProducerMessage(records, "passthrough")
```
But `TransactionalProducerMessage.apply` doesn't compile because `G[_]` is set to `Id` from `CommittableProducerRecords.one`.

I'm not sure if it's worth adding another type parameter and dealing with the `G[H[ProducerRecord[K, V]]` problem, but I do think it makes `CommittableProducerRecords.one` much less useful. What do you think?

vlovgr: It's a good point, and I thought about it while writing the implementation as well. I've pushed https://github.com/ovotech/fs2-kafka/pull/130/commits/88d17844ada00a6f9fdc3888fc29ccf37530b8ce which solves this by converting `H` to `G` using `Monad[G]`, `MonoidK[G]`, and `Foldable[H]`. It's not the most performant solution, but given you usually produce very few (~1) record per offset, I think the overhead isn't too bad here. It does prevent you from using 'non-empty' `G` collections though, due to the `MonoidK` requirement.

danxmoran:
⚠️ Scope creep ⚠️

I've been working to get an 0.19.x-compatible version of this PR set up in my project at work as a stop-gap while we work out the rough edges of the API here. The first problem I hit was picking a transactional ID. Alpakka asks for a user-defined ID, but I found that Kafka Streams auto-generates IDs for users. When I dug into the implementation I found that Kafka Streams [uses a unique transactional ID per "task" (topic/partition)](https://github.com/axbaretto/kafka/blob/fe51708ade3cdf4fe9640c205c66e3dd1a110062/streams/src/main/java/org/apache/kafka/streams/processor/TaskId.java). Other projects like [spring-kafka](https://github.com/spring-projects/spring-kafka/issues/800) have followed the same pattern. From what I've read, it's effectively the only way to ensure messages don't get double-processed when partitions get rebalanced.

Do you think this should change our design here at all? I could see a setup where we remove the `transactionalId` field from `ProducerSettings` and have `TransactionalKafkaProducer` manage the lifecycles of several "raw" Kafka producers under-the-hood. It'd be more complicated, but I think a big win. Without baked-in support, users who care about zombie-fencing guarantees will need to spin up a fresh `TransactionalKafkaProducer` every time they want to publish a new message.

vlovgr:
If I understand correctly, `TransactionalKafkaProducer` in its current form is only really useful if you have a single instance in the consumer group. With multiple instances, you need a stable `transactional.id` across instances per topic-partition, which is not trivial to implement yourself, and would fit better in the library. I definitely agree this is something we should handle in the library.

The mentioned topic-partition strategy also means one producer per topic-partition, right? That might be a bit much in the single instance scenario, but for multiple instances it sounds acceptable. Perhaps this is even behaviour we can toggle in `ProducerSettings`, while the remaining API stays as is? If you agree, then we can finish this up and merge it, and then do the remaining work in a separate pull request.

(I think the most tricky part of the producer-per-topic-partition is managing creating and closing producers as partitions are assigned and revoked. Maybe we could do something clever with rebalance listeners to get this working nicely.)

danxmoran:
:+1: I'm ok with this merging and setting up a new PR for improvements. Thanks for all the help and review! Is there anything I can do to help with adding docs & tests on this one?

vlovgr:
Great! If there's anything you feel is missing in terms of docs or tests, then feel free to add it. Otherwise, I'll just have a final look through this tomorrow and then merge it. 👍

src/main/scala/fs2/kafka/TransactionalProducerMessage.scala
vlovgr: I ended up changing `TransactionalProducerMessage` to only support `Chunk` for `G`, which solves the performance concerns. `Chunk` is going to be the the most used, if not only, type for `G` in practice, so think it's a reasonable restriction.

vlovgr:
Thanks a lot for this @danxmoran! 👍

PR #138, opened 2019-05-29
Add initial consumer documentation

vlovgr:


docs/src/main/mdoc/consumers.md
```diff
@@ -0,0 +1,314 @@
+---
+id: consumers
+title: Consumers
+---
+
+Consumers support subscribing to topics, record streaming and deserialization, as well as miscellaneous utility functionality, such as seeking to offsets, or checking what the end offsets are for a topic. Consumers make use of the [Java Kafka consumer][java-kafka-consumer], which becomes especially important for [settings](#settings). For consumer implementation details, refer to the [technical details](technical-details.md) section.
```
rmorais: why is it especially important for settings?

docs/src/main/mdoc/consumers.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+If the deserializer only needs access to the bytes, like in the case above, use `lift`.
+
+```scala mdoc:silent
+Deserializer.lift(bytes => IO.pure(bytes.dropWhile(_ == 0)))
+```
+
+To support different deserializers for different topics, use `topic`.
```
rmorais: Maybe be more explicit that we are pattern matching on topic names? 
Something like:
"To support different deserializers for different topics, use `topic`, which will allow to pattern match on the topic names."

docs/src/main/mdoc/consumers.md
```diff
@@ -0,0 +1,314 @@
...
+  headers("format").map(_.as[String]) match {
+    case Some("string") => Deserializer[IO, String]
+    case Some("int")    => Deserializer[IO, Int].map(_.toString)
+    case Some(format)   => Deserializer.failWith(s"unknown format: $format")
+    case None           => Deserializer.failWith("format header is missing")
+  }
+}
+```
+
+In the example above, `failWith` raises a [`SerializationException`][serializationexception] with the provided message.
```
rmorais: ~Shouldn't this be a `DeserializationException`?~
I can see from the source code below that the documentation is correct but this raises another question.
Given that in Kafka land, `Serialization` and `Deserialization` are distinct concepts, why have only one type of exception? Why not have a `DeserializationException` too?

docs/src/main/mdoc/consumers.md
```diff
@@ -0,0 +1,314 @@
...
+Deserializer.delegate[IO, String] {
+  new KafkaDeserializer[String] {
+    def deserialize(topic: String, data: Array[Byte]): String = new String(data)
+    def configure(configs: java.util.Map[String, _], isKey: Boolean): Unit = ()
+    def close(): Unit = ()
+  }
+}
+```
+
+If the deserializer performs _side effects_, follow with `suspend` to capture them properly.
```
rmorais: What will happen if we don't follow with `suspend` event though the serializer performs _side effects_?

docs/src/main/mdoc/consumers.md
```diff
@@ -0,0 +1,314 @@
...
+
+- `withCloseTimeout` controls the timeout when waiting for consumer shutdown. Default is 20 seconds.
+
+- `withCommitRecovery` defines how offset commit exceptions are recovered. See [`CommitRecovery.Default`][commitrecovery-default].
+
+- `withCommitTimeout` sets the timeout for offset commits. Default is 15 seconds.
+
+- `withCreateConsumer` changes how the underlying Java Kafka consumer is created. The default merely creates a Java `KafkaConsumer` instance using set properties, but this function allows overriding the behaviour for e.g. testing purposes.
+
+- `withExecutionContext` sets the `ExecutionContext` on which blocking Java Kafka consumer functions are executed. Unless specified, a default fixed single-thread pool is created as part of consumer initialization, with the thread name using the `fs2-kafka-consumer` prefix.
```
rmorais: for my curiosity, why does it use a fixed single-thread for blocking IO? I though the general rules was to use a cached thread pool for blocking IO? Is it because of order guarantees?

docs/src/main/mdoc/consumers.md
```diff
@@ -0,0 +1,314 @@
...
+    stream.compile.drain.as(ExitCode.Success)
+  }
+}
+```
+
+Note that only automatic partition assignment is supported. Like in the [consumer creation](#consumer-creation) section, the example above only creates a consumer (guaranteeing resource cleanup) and subscribes to a topic. No records are yet streamed from the topic, for which we'll have to use `stream` or `partitionedStream`.
+
+## Record Streaming
+
+Once subscribed to at least one topic, we can use `stream` for a `Stream` of [`CommittableMessage`][committablemessage]s. Messages each contain a deserialized [`ConsumerRecord`][consumerrecord], as well as a [`CommittableOffset`][committableoffset] for managing [offset commits](#offset-commits). Streams guarantee records in topic-partition order, but not ordering across partitions. This is the same ordering guarantee that Kafka provides.
```
rmorais: > Messages each contain a deserialized

I think this would read better as :
"Each message contains a deserialized"

docs/src/main/mdoc/consumers.md
vlovgr: Because settings (`ConsumerSettings`) include settings for the Java Kafka consumer.

docs/src/main/mdoc/consumers.md
vlovgr: I went with:
> To support different deserializers for different topics, use `topic` to pattern match on the topic name.

docs/src/main/mdoc/consumers.md
vlovgr: It's because _serialization_ is ambiguous. The term can refer to serialization as a concept (i.e. both `A => Array[Byte]` and `Array[Byte] => A`), or the serialization part specifically (i.e. `A => Array[Byte]`) . I do agree we should probably separate them though, so have introduced `DeserializationException`.

docs/src/main/mdoc/consumers.md
vlovgr: It should mainly impact errors, i.e. depending on the effect type you use, errors will either be handled anyway (with e.g. `IO` and `Task`) or treated as bug (with e.g. `ZIO`), resulting in immediate shutdown.

It's the same situation as `IO.pure(throw new RuntimeException)`, except we're in a `flatMap`-chain, meaning `IO` and `Task` will catch the error anyway, while `ZIO` won't. I'd like us to not describe how deserializers are used for the user though, so will leave it for now.

docs/src/main/mdoc/consumers.md
vlovgr: It's because we know we need exactly one thread for the blocking IO. If the user already has a cached thread pool for blocking IO, they can set to use it with `withExecutionContext`.

docs/src/main/mdoc/consumers.md
rmorais: sounds good to me!!

PR #158, opened 2019-07-17
Expose alterConfigs functionality

joceron:
I don't know if I should create this PR just like that, but I needed to update the configurations of a topic after it was created, and I saw that `org.apache.kafka.clients.admin.AdminClient` already has that functionality, but `fs2.kafka.KafkaAdminClient` doesn't expose it. So... all the feedback in the world is welcome


vlovgr:
Thanks for this @joceron!

PR #160, opened 2019-07-23
Add support for auto-assignment of transactional IDs based on consumer rebalances

danxmoran:
_Very_ rough-draft, but wanted to get some eyes on the general approach.

@vlovgr can you take a look, when you get a minute?

### DIFF
--- a/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
+++ b/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
@@ -4,11 +4,12 @@ import cats.effect.IO
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import fs2.Stream
-import fs2.concurrent.Queue
+import fs2.concurrent.{NoneTerminatedQueue, Queue}
 import net.manub.embeddedkafka.EmbeddedKafkaConfig
 import org.apache.kafka.clients.consumer.NoOffsetForPartitionException
 import org.apache.kafka.common.TopicPartition
 
+import scala.collection.immutable.SortedSet
 import scala.concurrent.duration._
 
 final class KafkaConsumerSpec extends BaseKafkaSpec {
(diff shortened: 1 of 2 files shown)


vlovgr:
@danxmoran what do you think about linking the producer to the consumer by simply passing the `KafkaConsumer` instance to the function creating the producer? The producer could then register the listeners it needs without having to change `ConsumerSettings` or create the consumer itself.

danxmoran:
@vlovgr that sounds OK to me! Would the producer-builder still return a new `KafkaConsumer` copy in that case, where the new instance has the listeners registered? Or would it update the state of the existing instance?

vlovgr:
@danxmoran I would suggest to register listeners on the existing instance using `Request.Assignment`, similar to how we do it in `partitionedStream`. We might need to add a new `assignment` overload to `KafkaConsumer` for exposing the ability to add rebalance listeners.

https://github.com/ovotech/fs2-kafka/blob/5256957c03fe91581a57469e31c46aa743dda545/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala#L420-L433

danxmoran:
@vlovgr I was thinking about your suggestion for an `assignment` overload with listeners, and wondered if this would be an even-better alternative for a new method:
```scala
def assignmentStream: Stream[F, SortedSet[TopicPartition]]
```
What do you think? The stream would emit on every rebalance, but the details of how the rebalance maps to a `SortedSet` could be fully encapsulated by the `KafkaConsumer`. It seems like a closer match to the other APIs on the consumer, and it'll save us from needing to encode too many details about the `KafkaConsumer` in the `TransactionalKafkaProducer` setup.

vlovgr:
@danxmoran I like the idea! The only downside I can think of is that users of `assignmentStream` will have to work out what partitions have been assigned or revoked (checking differences between previous and current sets). If this should prove challenging, we could introduce an `Assignment` wrapper which includes both the current partitions and what was revoked/added.

danxmoran:
I took a stab at a naive implementation of `assignmentStream`, with a basic test. It's not behaving as expected, but I can't tell if it's because of a problem with Kafka or (probably) a problem with how I'm trying to set up concurrent operations. Can you take a look and let me know if anything obvious jumps out, @vlovgr?

modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -352,4 +353,39 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
...
+          queue <- Stream.eval(Queue.noneTerminated[IO, SortedSet[TopicPartition]])
+          _ <- consumer(queue)
+          _ <- Stream.eval(IO.sleep(5.seconds))
+          _ <- consumer(queue)
+          _ <- Stream.eval(IO.sleep(5.seconds))
+          _ <- Stream.eval(queue.enqueue1(None))
+          allUpdates <- Stream.eval(queue.dequeue.compile.toList)
+          _ <- Stream.eval(IO(assert {
+            println(allUpdates)
+            allUpdates.length == 3
```
danxmoran: I'm thinking 3 because I'd expect:
1. Initial assignment of first consumer
2. Initial assignment of second consumer
3. Rebalanced set from first consumer, after the second joins

What I'm seeing instead is a list of two empty sets. Something is _really_ off because the `println`s in the main code show that we're hitting the `Init` and `Update` branches for both consumers, so we should be enqueuing 4 things.

vlovgr:
@danxmoran It's probably a good idea to move `assignmentStream` to a separate pull request. The issue you're facing happens because `assignment` is used in the callback. The callbacks are run synchronously, waiting for the effects to finish before returning (which is the `ConsumerRebalanceListener` contract).

https://github.com/ovotech/fs2-kafka/blob/5256957c03fe91581a57469e31c46aa743dda545/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala#L79-L86

While the callback effects are being run, no new requests are handled by the actor, and as such, the `assignment` request won't be completed. A solution to this is to not call `assignment`, but instead store the currently assigned partitions in a `Ref[F, SortedSet[TopicPartition]]`, and update the `Ref` + emit current assignment when rebalance happens.

danxmoran:
Ack, ok. Was worried that might be the case. I'll close this for now and reopen a PR for just the `assignmentStream`, then revisit the auto-rebalancing producers after that.

Thanks for all the help @vlovgr! I'll get there eventually 😄

PR #169, opened 2019-08-07
Add assignmentStream to KafkaConsumer.

danxmoran:
Split from #160, to support auto-rebalancing transactional producers.


vlovgr:
Thanks @danxmoran! 

There's one race condition: there's a chance that rebalance happens between asking for the initial `assignment` and registering the rebalance listeners. In your version, you could even remove the initial `assignment` and update the `Ref` with the `updated` assignment from when registering the rebalance listeners. However, there's still a chance that rebalance happens before you're able to update the `Ref` with the initial assignment, so we need to handle that case as well.

I've pushed a version which fixes the race condition, please have a look.

Some additional notes:
- The reason rebalance listeners are registered as part of `assignment` (and not a separate operation) is to avoid deadlock. Here, we semantically block in the listeners until we've had a chance to set the initial assignments. Since we return the initial assignment together with registration, we're always able to set the initial assignment and unblock the listeners. If assignment happened after listeners were registered, we might never get a reply if the `KafkaConsumerActor` is busy waiting in listeners.
- We face the same issue in `partitionedStream`, except there the order doesn't matter for assignments, and we handle revocations by checking whether the requested partition is still assigned on every fetch (instead of using the `onRevoked` listener).

danxmoran:
Thanks @vlovgr, it looks good to me! I just added a comment capturing your explanation about the race condition, for anyone else who reads through the code.

Issue #132, opened 2019-05-14
Add automatic management of transactional.id to TransactionalKafkaProducer


danxmoran:
Follow-up from #128 / #130.

Quoting the PR:
> I found that Kafka Streams uses a unique transactional ID per "task" (topic/partition). Other projects like spring-kafka have followed the same pattern. From what I've read, it's effectively the only way to ensure messages don't get double-processed when partitions get rebalanced.

> If I understand correctly, TransactionalKafkaProducer in its current form is only really useful if you have a single instance in the consumer group. With multiple instances, you need a stable transactional.id across instances per topic-partition, which is not trivial to implement yourself, and would fit better in the library. I definitely agree this is something we should handle in the library.

> The mentioned topic-partition strategy also means one producer per topic-partition, right? That might be a bit much in the single instance scenario, but for multiple instances it sounds acceptable. Perhaps this is even behaviour we can toggle in ProducerSettings, while the remaining API stays as is? If you agree, then we can finish this up and merge it, and then do the remaining work in a separate pull request.

> (I think the most tricky part of the producer-per-topic-partition is managing creating and closing producers as partitions are assigned and revoked. Maybe we could do something clever with rebalance listeners to get this working nicely.)

danxmoran:
Hey @vlovgr, just a heads-up that I'm going to try to get a PR up for this in the next few days. I'm going to try to hook up the rebalance listeners as you suggested.

PR #174, opened 2019-08-09
Prep for adding auto-rebalancing transactional producer.

danxmoran:
Part of #132.

I've got a WIP branch implementing the auto-rebalancing variant, but I'm still working out the concurrency details. The diff is big enough now that I thought it'd be worth pushing the pieces that refactor existing code.

I'll comment inline with explanations.


modules/core/src/main/scala/fs2/kafka/CommittableProducerRecords.scala
```diff
@@ -34,26 +35,21 @@ import fs2.kafka.internal.syntax._
   * - `CommittableProducerRecords#one` to produce exactly one record within
   * the same transaction as the offset is committed.
   */
-sealed abstract class CommittableProducerRecords[F[_], G[+_], +K, +V] {
```
danxmoran: Trying to support `G[_]` here is a major pain when trying to batch records into transactions. The value-add seems to be allowing users to build instances of this class from semi-arbitrary collections of records. We can get the same value by keeping `G[_]` in the static builder methods, but then convert to `Chunk` right away to make downstream processing simpler.

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -42,90 +42,124 @@ abstract class TransactionalKafkaProducer[F[_], K, V] {
...
-                          )
-                          .map(_.sequence)
-                          .flatTap { _ =>
-                            F.delay {
-                              producer.sendOffsetsToTransaction(
-                                batch.offsets.asJava,
-                                groupId
-                              )
+                    consumerGroupId.flatMap { groupId =>
+                      records.flatTraverse { committable =>
```
danxmoran: This is one block where replacing `G[_]` with `Chunk` pays off, because it makes interleaving the various `traverse` + `flatten` combinators much easier.

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -42,90 +42,124 @@ abstract class TransactionalKafkaProducer[F[_], K, V] {
...
+      * single instance in the consumer group. True exactly-once processing is
+      * only guaranteed by running separate producers with unique transactional
+      * IDs per topic/partition processed by the group.
+      */
+    def withConstantId: Resource[F, TransactionalKafkaProducer[F, K, V]] =
+      resource { blocker =>
+        Resource
+          .make(settings.createProducer)(close(blocker))
+          .evalTap(init(blocker))
+          .map { producer => records =>
```
danxmoran: This is the other spot where removing `G[_]` pays off. Without the method-level type parameter we can pass the grouping logic around as a simple function. If we keep the `G[_]`, we have to introduce a new `trait` with a generic method and construct that here instead.

modules/core/src/main/scala/fs2/kafka/TransactionalProducerRecords.scala
```diff
@@ -37,20 +37,20 @@ import fs2.kafka.internal.syntax._
   * commit exactly one offset, then emit a [[ProducerResult]] with the
   * results and specified passthrough value.
   */
-sealed abstract class TransactionalProducerRecords[F[_], G[+_], +K, +V, +P] {
```
danxmoran: As far as I can tell, the only point of this `G[_]` was to pass through the type of the wrapped `records` field.

modules/core/src/main/scala/fs2/kafka/package.scala
```diff
@@ -257,7 +257,7 @@ package object kafka {
     implicit F: ConcurrentEffect[F],
     context: ContextShift[F]
   ): Resource[F, TransactionalKafkaProducer[F, K, V]] =
-    TransactionalKafkaProducer.resource(settings)
+    new TransactionalKafkaProducer.Builder[F, K, V](settings).withConstantId
```
danxmoran: I expect we'll need to rethink the syntax helpers

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -42,90 +42,124 @@ abstract class TransactionalKafkaProducer[F[_], K, V] {
...
+    /**
+      * Construct a [[TransactionalKafkaProducer]] which uses a constant ID
+      * for all transactions.
+      *
+      * NOTE: This mode only guarantees exactly-once processing if you have a
+      * single instance in the consumer group. True exactly-once processing is
+      * only guaranteed by running separate producers with unique transactional
+      * IDs per topic/partition processed by the group.
+      */
+    def withConstantId: Resource[F, TransactionalKafkaProducer[F, K, V]] =
```
danxmoran: @vlovgr I _really_ want to make this take a `String` parameter to use as the constant ID and have this method inject the appropriate config, to entirely dodge the chance of a runtime error from passing in settings without a transactional ID, but then there isn't much point to having `withTransactionalId` on `ProducerSettings`. What do you think?

It makes even more sense in the dynamic case, where we need an arbitrary prefix for the rebalancing transactional IDs. Pulling the "transactional ID" out of settings and using it as the prefix works, but feels like it could be confusing to users. It also forces our code to deal with the error case when there's no ID present in config, vs. just throwing it at Kafka and catching any errors that bubble up.

modules/core/src/main/scala/fs2/kafka/CommittableProducerRecords.scala
vlovgr: Sounds good! 👍

modules/core/src/main/scala/fs2/kafka/package.scala
vlovgr: Yes, let's worry about that later on though. :)

modules/core/src/main/scala/fs2/kafka/CommittableProducerRecords.scala
```diff
@@ -67,8 +63,17 @@ object CommittableProducerRecords {
   def apply[F[_], G[+_], K, V](
     records: G[ProducerRecord[K, V]],
     offset: CommittableOffset[F]
-  )(implicit G: Foldable[G]): CommittableProducerRecords[F, G, K, V] =
-    new CommittableProducerRecordsImpl(records, offset, G)
+  )(implicit G: Foldable[G]): CommittableProducerRecords[F, K, V] = {
+    val buf = new mutable.ArrayBuffer[ProducerRecord[K, V]]()
```
vlovgr: I think it's worth initializing this to `records.size`, as many `Foldable` have constant-time `size`.

vlovgr: We can also specialize with `Chunk.empty` and `Chunk.singleton` on sizes 0 and 1, respectively. Especially since `one` uses `apply` (also possible to change `one` to `new CommittableProducerRecordsImpl(Chunk.singleton(record), offset)`).

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -42,90 +42,124 @@ abstract class TransactionalKafkaProducer[F[_], K, V] {
...
+        Resource.liftF(settings.valueSerializer).flatMap { valueSerializer =>
+          settings.blocker.fold(Blockers.producer)(Resource.pure[F, Blocker]).flatMap { blocker =>
+            buildGrouper(blocker).map { grouper =>
+              new TransactionalKafkaProducer[F, K, V] {
+                override def produce[P](
+                  records: TransactionalProducerRecords[F, K, V, P]
+                ): F[ProducerResult[Chunk, K, V, P]] =
+                  grouper(records.records)
+                    .flatMap { recordGroups =>
+                      recordGroups.flatTraverse {
```
vlovgr: Should these run in parallel rather than sequentially?

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -42,90 +42,124 @@ abstract class TransactionalKafkaProducer[F[_], K, V] {
...
-                "TransactionalKafkaProducer$" + System.identityHashCode(this)
+                override def toString: String =
+                  "TransactionalKafkaProducer$" + System.identityHashCode(this)
+              }
             }
           }
+        }
+      }
+
+    private def init(blocker: Blocker)(producer: KafkaByteProducer): F[Unit] =
```
vlovgr: You might want to make use of [`WithProducer`](https://github.com/ovotech/fs2-kafka/blob/master/modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala) in `TransactionalKafkaProducer`.

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
vlovgr: Agree, let's move transaction-specific configuration to these functions, rather than the Settings.

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
danxmoran: :+1: I think they should, we just need the extra type parameter.

modules/core/src/main/scala/fs2/kafka/CommittableProducerRecords.scala
danxmoran: `size` on `Foldable` returns `Long`, but `ArrayBuffer`'s constructor expects `Int`. What's the best way to safely handle the overflow?

vlovgr: I'd say it's unlikely you would want to produce more than `Int.MaxValue` records at a time, so `.toInt` sounds reasonable to me.

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
danxmoran: I saw that, but I think the way it's set up will prevent manual setup / close management for `KafkaByteProducer`s that we'll need for the rebalancing case.

danxmoran:
@vlovgr the last commit is a (pretty aggressive) refactor addressing the review comments. Moving the transaction parameters out of `ProducerSettings` broke all of the existing syntax helpers, so I deleted them for now. I'm 👍 to restoring them before this PR merges, I just wanted to reduce the noise while iterating on the core logic.

I also added a new test case for transaction timeouts. While working on it, I found that transaction timeouts end up raising `ProducerFencedException` instances. The doc for that error is:
```java
/**
 * This fatal exception indicates that another producer with the same <code>transactional.id</code> has been
 * started. It is only possible to have one producer instance with a <code>transactional.id</code> at any
 * given time, and the latest one to be started "fences" the previous instances so that they can no longer
 * make transactional requests. When you encounter this exception, you must close the producer instance.
 */
```
Emphasis _When you encounter this exception, you must close the producer instance_. Once I was poking around the javadocs I saw the same message listed on nearly every method as the only "safe" way to recover from an error mid-transaction.

Now that we've introduced an abstraction layer between `TransactionalKafkaProducer` and the underlying `KafkaByteProducer`, do you think it's worth trying to engineer an auto-restart of the underlying producer on errors? Or is it fine to bubble up errors and leave it to users to close & restart when desired?

modules/core/src/test/scala/fs2/kafka/BaseKafkaSpec.scala
```diff
@@ -51,7 +56,10 @@ abstract class BaseKafkaSpec extends BaseAsyncSpec with EmbeddedKafka {
   final def withKafka[A](f: (EmbeddedKafkaConfig, String) => A): A =
     withRunningKafkaOnFoundPort(
       EmbeddedKafkaConfig(
-        customBrokerProperties = Map("transaction.state.log.replication.factor" -> "1")
+        customBrokerProperties = Map(
+          "transaction.state.log.replication.factor" -> "1",
+          "transaction.abort.timed.out.transaction.cleanup.interval.ms" -> transactionTimeoutInterval.toMillis.toString
```
danxmoran: This is needed to force the broker to actually notice when open transactions time out.

vlovgr:
> do you think it's worth trying to engineer an auto-restart of the underlying producer on errors? Or is it fine to bubble up errors and leave it to users to close & restart when desired?

I think it would be useful to handle errors, similarly to how [`CommitRecovery`](https://github.com/ovotech/fs2-kafka/blob/master/modules/core/src/main/scala/fs2/kafka/CommitRecovery.scala) handles commit errors by default (although configurable). I think that can be implemented in a separate pull request though.

modules/core/src/main/scala/fs2/kafka/CommittableProducerRecords.scala
```diff
@@ -67,8 +63,25 @@ object CommittableProducerRecords {
...
-    new CommittableProducerRecordsImpl(records, offset, G)
+  )(implicit G: Foldable[G]): CommittableProducerRecords[F, K, V] = {
+    val numRecords = G.size(records)
+    val chunk = if (numRecords <= 1) {
+      G.get(records)(0) match {
+        case None         => Chunk.empty[ProducerRecord[K, V]]
+        case Some(record) => Chunk.singleton(record)
+      }
+    } else {
+      val buf = new mutable.ArrayBuffer[ProducerRecord[K, V]](G.size(records).toInt)
```
vlovgr: `G.size(records).toInt` -> `numRecords.toInt`

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -42,90 +45,142 @@ abstract class TransactionalKafkaProducer[F[_], K, V] {
...
+    type Records = CommittableProducerRecords[F, K, V]
+    type RecordsChunk = Chunk[Records]
+    type RecordGrouper = RecordsChunk => F[Chunk[(KafkaByteProducer, RecordsChunk)]]
+
+    private[this] def resource(
+      buildGrouper: Blocker => Resource[F, RecordGrouper]
+    ): Resource[F, TransactionalKafkaProducer[F, K, V]] =
+      Resource.liftF(settings.keySerializer).flatMap { keySerializer =>
+        Resource.liftF(settings.valueSerializer).flatMap { valueSerializer =>
+          settings.blocker.fold(Blockers.producer)(Resource.pure[F, Blocker]).flatMap { blocker =>
```
vlovgr: We should probably introduce a `Blocker.transactionalProducer`, which uses a cached thread-pool rather than a single executor thread pool, since we'll want multiple transactions to run in parallel. `Blockers.producer` will only allow one transaction to run at-a-time.

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
vlovgr: > manual setup / close management

We'll need to have a `bracket` in there somewhere to ensure we `close` no matter what (e.g. on errors). But this can be done as a `Stream` with `Stream.bracket`.

danxmoran:
@vlovgr do you want syntax bike-shedding to live in this PR, or would you rather wait until this and the rebalancing functionality are in? I have ideas, but I appreciate wanting to keep the PRs focused too 😄

If we're waiting on the syntax, what's the best way to deal with the mdoc errors?

vlovgr:
@danxmoran Yeah, let's do the syntax in a separate pull request. You can disable mdoc temporarily for a block by changing `scala mdoc:<modifier>`  to `scala` for the relevant blocks. Maybe leave a TODO so we don't forget to add it back.

Is this ready from your perspective? If so, I'll try to have a final lookthrough today or tomorrow, tidy up any last bits, and get it merged.

danxmoran:
@vlovgr I think this is ready, apart from the mdoc

vlovgr:
@danxmoran Sorry it's taken me a while longer to look through this properly. Happy to merge as-is. Thanks! 👍

PR #177, opened 2019-08-13
Move isKey from AvroSettings to Record

vlovgr:
`AvroSettings[F[_], A]` has caused confusion, since `A` is a phantom type. The sole reason for `A` is to be able to have type-specific settings. It turns out, the only type-specific setting is whether `A` is used for record keys or values (configurable using `withIsKey(Boolean)`). It's also an annoyance to have to specify `withIsKey` when it's clear which serializer is used for keys and which is used for values.

This pull request changes to `AvroSettings[F[_]]` and introduces serializers and deserializers which can vary depending on whether they're used for record keys or values (in the form of `Serializer.Record` and `Deserializer.Record`). This way, `withIsKey` is no longer needed and `isKey` will be set correctly automatically, and `AvroSettings` is no longer type-specific, avoiding confusion.

For comparison: [before](https://ovotech.github.io/fs2-kafka/docs/modules#vulcan) and [after](https://github.com/ovotech/fs2-kafka/blob/35ec7396e904e60a0efacebdce9527fd3d61902c/docs/src/main/mdoc/modules.md) the changes in this pull request.


modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
```diff
@@ -384,62 +383,51 @@ object ProducerSettings {
...
-    * are wrapped in a creation effect.
-    */
   def apply[F[_], K, V](
-    keySerializer: F[Serializer[F, K]],
-    valueSerializer: F[Serializer[F, V]]
+    keySerializer: Serializer.Record[F, K],
+    valueSerializer: Serializer.Record[F, V]
   )(implicit F: Sync[F]): ProducerSettings[F, K, V] =
-    create(keySerializer, valueSerializer)
+    create(
```
julien-truffaut: I find this quite confusing to take two `Serializer.Record` but only use the key / value side of it

julien-truffaut:
Same for me, looks good. Thanks @vlovgr

modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
vlovgr: It's valid though, since `Serializer.Record` provides two serializers.

For example:

```scala
val serializer = avroSerializer[String].using(avroSettings)

ProducerSettings(
  keySerializer = serializer,
  valueSerializer = serializer
)
```

would use both parts, or the following:

```scala
ProducerSettings(
  keySerializer = avroSerializer[String].using(avroSettings),
  valueSerializer = avroSerializer[Person].using(avroSettings)
)
```

where you would use Avro for both serializers, but only the `forKey` part for `String` and the `forValue` part for `Person`.

vlovgr:
> Maybe all the overloads can cause similar problems for other users, but I'm not sure

I'm hoping not, but let's see. We can always do more changes if it turns out to be an issue.

Thanks @SystemFw @julien-truffaut for reviewing!

PR #182, opened 2019-08-19
WIP First pass at rebalancing producer.

danxmoran:
Opening as a draft because there's one failing test that I'm not sure how to handle.

### DIFF
--- a/modules/core/src/test/scala/fs2/kafka/TransactionalKafkaProducerSpec.scala
+++ b/modules/core/src/test/scala/fs2/kafka/TransactionalKafkaProducerSpec.scala
@@ -6,6 +6,7 @@ import cats.data.NonEmptyList
 import cats.effect.IO
 import cats.implicits._
 import fs2.{Chunk, Stream}
+import fs2.kafka.internal.instances._
 import fs2.kafka.internal.converters.collection._
 import net.manub.embeddedkafka.EmbeddedKafkaConfig
 import org.apache.kafka.clients.consumer.{ConsumerConfig, OffsetAndMetadata}
@@ -15,6 +16,7 @@ import org.apache.kafka.common.serialization.ByteArraySerializer
 import org.scalatest.{EitherValues, OptionValues}
 
 import scala.concurrent.duration._
+import scala.collection.immutable.SortedSet
 
 class TransactionalKafkaProducerSpec extends BaseKafkaSpec with OptionValues with EitherValues {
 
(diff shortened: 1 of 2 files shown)


modules/core/src/test/scala/fs2/kafka/TransactionalKafkaProducerSpec.scala
```diff
@@ -267,4 +269,204 @@ class TransactionalKafkaProducerSpec extends BaseKafkaSpec with OptionValues wit
...
+            zooKeeperPort = config.zooKeeperPort,
+            customConsumerProperties =
+              Map(ConsumerConfig.ISOLATION_LEVEL_CONFIG -> "read_committed")
+          )
+          Either.catchNonFatal(consumeFirstKeyedMessageFrom[String, String](topic))
+        }
+
+        // FIXME: This currently fails because committing offsets only fails in the producer
+        // for one of the three topic/partitions. The other t/p's successfully commit messages,
+        // but we don't currently have a mechanism to signal what worked and what didn't.
```
danxmoran: @vlovgr here's a case that I hadn't fully thought through. Now that a `TransactionalKafkaProducer` can wrap multiple `KafkaByteProducer`s, producing a single `TransactionalProducerRecords` isn't atomic any more. As-is, the only way to safely recover from a transaction failure would be to restart the whole stream and re-process any messages that weren't committed. We might be able to do better by raising some custom error containing all the records that _were_ successfully committed, but I'm not sure it's safe / worth it. What do you think?

vlovgr: > Now that a `TransactionalKafkaProducer` can wrap multiple `KafkaByteProducer`s, producing a single `TransactionalProducerRecords` isn't atomic any more.

I think users might find it confusing that `TransactionalProducerRecords` is no longer atomic with this construct, but atomic if using a single producer. It might be worth considering an alternative construct or way to communicate what is and isn't atomic. If nothing else, we should try to cleary document this.

> As-is, the only way to safely recover from a transaction failure would be to restart the whole stream and re-process any messages that weren't committed. We might be able to do better by raising some custom error containing all the records that were successfully committed, but I'm not sure it's safe / worth it. What do you think?

Raising the error and letting the user decide sounds acceptable for a first pass. We can explore recovery options (similar to `CommitRecovery`) in a subsequent pull request. Also, the standard producer doesn't recover from any errors currently.

danxmoran: One option for dealing with the atomicity problem would be to drop `TransactionalProducerRecords`, and move `passthrough` onto `CommittableProducerRecords`. In https://github.com/ovotech/fs2-kafka/pull/130#discussion_r283091204 it looks like we decided to keep both so users could produce to multiple topic/partitions in a single transaction, but now there isn't really a single transaction so I'm not sure if the abstraction holds up.

danxmoran: I wonder if I'm attacking this at the wrong layer. Instead of having a single `TransactionalKafkaProducer` wrap multiple `KafkaByteProducer`s, we could have a 1:1 mapping and then a higher-level helper to manage spinning the `TransactionalKafkaProducer` instances up/down. The helper could be a:
```scala
Pipe[F, CommittableProducerRecords, Chunk[ProducerResult]]
```
Where each emitted `Chunk` originates a single successfully-completed transaction. Batching for transactions could be tweaked using similar arguments as `groupWithin`. If the transaction for a particular topic/partition fails, the pipe would finish emitting the results of any other in-flight transactions, then fail with the error.

What do you think?

vlovgr: That sounds like an interesting idea to explore. I agree chunking probably needs to be handled outside, as one needs to manually determine what constitutes a transaction. This also fits well with the existing `produce` pipes.

vlovgr:
> I'm closing this to reduce the noise while I'm working on the new idea. It might take awhile, but I do still plan on contributing this feature if nobody else gets to it first

It's a great addition and am happy to assist in getting it added. :)

> Would you like me to open a PR reverting everything but those two things, so you're clear to cut 0.20.0-final if you think it's time but I haven't opened an alternative PR yet?

Yes, that would be great!

> Side question: I see there are still some `G[_]` bounds floating in the non-transactional types. Would it make sense to replace them all with `Chunk` for consistency & efficiency?

Yes, sounds good.

PR #187, opened 2019-08-27
Use Chunk in ProducerRecords / ProducerResult.

danxmoran:
As proposed in https://github.com/ovotech/fs2-kafka/pull/182#issuecomment-525091340.


vlovgr:
Looks great, thanks! 👍

PR #193, opened 2019-08-29
Add more info to produce

SystemFw:
Very rough around the edges, but better than nothing for now. Feel free to edit of course.


vlovgr:
Thanks @SystemFw! I did some minor cleanup.

PR #265, opened 2019-11-23
Provide functions for creating KafkaProducer test instances

hmagrahi:


aartigao:
For this you can use [fs2-kafka-mock](https://index.scala-lang.org/jchapuis/fs2-kafka-mock).

PR #269, opened 2019-11-27
Proposal to alias `Serializer.Record`/`Deserializer.Record` as `RecordSerializer`/`RecordDeserializer`

bplommer:
I've found the current naming confusing, both when learning the API and when explaining it to others: the default assumption for a newcomer is that `Deserializer.Record` is either a subtype of `Deserializer`, or a kind of record that is used in combination with one. It's also awkward to say "deserializer dot record" out loud.

I think this change improves the situation by presenting these types as first-class elements of the API.


vlovgr:
Thanks @bplommer! Let's properly rename these for the next binary-breaking release (https://github.com/fd4s/fs2-kafka/issues/276).

PR #270, opened 2019-11-27
Add `apply` methods for `AvroSerializer`/`AvroDeserializer`

bplommer:
Particularly useful in IntelliJ, which will not auto-import the factory methods from the `fs2.kafka.vulcan` package object, creating temptation to import that object, which in turn breaks auto-import of anything from the root `vulcan` namespace.


vlovgr:
Thanks @bplommer!

PR #251, opened 2019-10-28
add ACLs describe, crete, delete method in KafkaAdminClient

novakov-alexey:
Closes #250

PR #273, opened 2019-12-05
Additional consumer functionality

vavr:
Partially resolves this issue: fd4s#94

Added some methods to a consumer.

These methods are just proxies to java's consumer:
* `def partitionsFor(topic: String): F[List[PartitionInfo]]` - list of partitions info for topic
* `def partitionsFor(topic: String, timeout: FiniteDuration)` - same but with timeout
* `def metrics: F[Map[MetricName, Metric]]` - get consumer metrics

These methods are working with the actor state. All `assign` methods set the actor state to `subscribed`.
* `def assign(topicPartitions: NonEmptySet[TopicPartition]): F[Unit]` - assign set of topic partitions
* `def assign(topic: String, partitions: NonEmptySet[Int]): F[Unit]` - assign partitions of one topic
* `def assign(topic: String): F[Unit]` - assign to all partitions of a given topic
* `def unsubscribe: F[Unit]` - unsubscribe from all topics and partitions assigned by `subscribe` or `assign`. Set `subscribed` state to false.

All methods have integrational tests.


vavr:
In this PR, I added some useful methods, but mima plugin does not pass binary compatibility checks.

`[error]  * abstract method unsubscribe()java.lang.Object in class fs2.kafka.KafkaConsumer is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaConsumer.unsubscribe")`

It looks very strange :)

@vlovgr Can you comment on this behavior somehow?

vavr:
I've found the PR https://github.com/fd4s/fs2-kafka/pull/251 with the same problem, and there is a solution like this:  https://github.com/fd4s/fs2-kafka/pull/251/commits/d1f17084a3f312c77f4bb147d45b4331ad3e3529#diff-fdc3abdfd754eeb24090dbd90aeec2ceR230-R233

I can add filters for new methods BUT! the best way is somehow to configure the mima to check only the back capability.

vlovgr:
Thanks @vavr! Adding mima exclusions is the right thing to do here. I pushed some very minor cleanup, this looks great, thanks.

PR #277, opened 2019-12-10
Convert Headers to use Map as underlying data structure

CremboC:
Currently `Headers` is based of a `Chain`. This we need to traverse the whole `Chain` to find values in it (`O(n)`). At a large scale, this becomes an expensive operation, especially in our case where we need to extract two headers which may or may not be there. The operation is clearly visible in the profiler.

This PR changes `Headers` to be based of `Map` instead, making lookups `O(1)`.

If you guys think this is useless, let me know! Perhaps `Headers` used to be based on `Map` but it was inefficient or wrong in some other way?

**Not addressed in this PR**:
The underlying Kafka `Headers` class actually allows to have multiple values for the same key. The current `fs2-kafka` `Headers` class doesn't really allow it because the main function is:
```scala
final def apply(key: String): Option[Header]
```
I believe this should be changed into either:
```scala
final def apply(key: String): List[Header]
```
or
```scala
final def apply(key: String): Option[NonEmptyList[Header]]
```
However this PR does not address that.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/Headers.scala
+++ b/modules/core/src/main/scala/fs2/kafka/Headers.scala
@@ -139,7 +139,7 @@ object Headers {
 
         override def headers(key: String): java.lang.Iterable[KafkaHeader] = {
           map.get(key) match {
-            case Some(value) => Iterable.single(value: KafkaHeader).asJava
+            case Some(value) => Iterable.apply(value: KafkaHeader).asJava
             case None => Iterable.empty[KafkaHeader].asJava
           }
         }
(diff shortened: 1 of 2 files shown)


vlovgr:
Thanks for this @CremboC! I believe the current `Headers` implementation was meant to be consistent with the Java Kafka one, specifically in terms of allowing multiple entries with the same key. I can't recall why this is allowed in the Java implementation, but if we can figure out there's no real reason for it, then I'm happy to go ahead with this. :+1:

CremboC:
I did some performance testing with a high load with the following properties:
- 90k messages per second from Kafka
- The headers are *always* present
- Looking for two headers (and I think there are 3 in the messages)

With that setup and this branch I got 1.41x better performance (observed in the profiler for that single function).

I would be curious to try another thing - use `Array` as the underlying data structure and traversing that. I'll test that and report back.

vlovgr:
From the [documentation](https://kafka.apache.org/24/javadoc/org/apache/kafka/connect/header/Headers.html) for the Java Kafka `Headers`:

> A mutable ordered collection of Header objects. Note that multiple headers may have the same key.

As such, I reckon we should allow multiple headers for the same key. I'm going to close this for now, but would definitely be great to explore performance improvements while still allowing multiple values for the same key.

CremboC:
@vlovgr that's fair! One quick (albeit tiny) optimisations would be to use an Array instead of `Chain`.

PR #282, opened 2019-12-14
Vulcan: Use writer schema from registry for deserialization

bplommer:
Currently, the deserializer passes the reader schema rather than the writer schema to Vulcan, which means you don't get validation at the top level against the writer schema.


modules/vulcan/src/test/scala/fs2/kafka/vulcan/PackageSpec.scala
```diff
@@ -29,6 +33,26 @@ final class PackageSpec extends AnyFunSpec {
...
+
+    it("should error when reader and writer schemas have mismatching logical types") {
+      (for {
+        serializer <- avroSerializer[Long].using(avroSettings).forValue
+        rawLong = 42L
+        serialized <- serializer.serialize("topic3", Headers.empty, rawLong)
+        deserializer <- avroDeserializer[Instant].using(avroSettings).forValue
+        deserialized <- deserializer.deserialize("topic3", Headers.empty, serialized).attempt
+      } yield assert(deserialized.isLeft)).unsafeRunSync
+    }
```
bplommer: This test fails before the changes in this PR.

modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroDeserializer.scala
```diff
@@ -29,13 +31,18 @@ final class AvroDeserializer[A] private[vulcan] (
...
-                codec.decode(deserializer.deserialize(topic, bytes, schema), schema) match {
-                  case Right(a)    => F.pure(a)
-                  case Left(error) => F.raiseError(error.throwable)
+        val createDeserializer: Boolean => F[Deserializer[F, A]] = isKey =>
+          settings.schemaRegistryClient.flatMap { schemaRegistryClient =>
+            settings.createAvroDeserializer(isKey).map { deserializer =>
+              Deserializer.instance { (topic, _, bytes) =>
+                F.suspend {
+                  val writerSchemaId = ByteBuffer.wrap(bytes).getInt(1) // skip magic byte
+                  val writerSchema = schemaRegistryClient.getById(writerSchemaId)
```
bplommer: It would be nice not to have to do this byte-level stuff to get the writer schema, but I don't think the schema registry API gives a way to avoid it.

vlovgr:
@bplommer I changed `createAvroDeserializer` to also return the `SchemaRegistryClient`, so we avoid creating a second client for every deserializer. (`createAvroSerializer` was also changed to maintain consistency.)

Issue #234, opened 2019-10-11
Provide functions for creating KafkaConsumer test instances


vlovgr:
It would be useful to provide functions for creating test versions of `KafkaConsumer`.

As a first step, the following function would be a useful addition.

```scala
object KafkaConsumer {
  def fromRecords[F[_], K, V](records: F[ConsumerRecord[K, V]])(
    implicit F: Foldable[F]
  ): KafkaConsumer[F, K, V] =
    ???
}
```

Following, it would be useful to be able to track what offsets have been committed.

lewisjkl:
I am starting to take a look at this. I have some ideas of how to take `F[ConsumerRecord[K, V]]` and translate them into a `Stream[F, Stream[F, CommittableConsumerRecord[F, K, V]]]` for the `partitionedStream` method. As long as you don't have any strong opinions about how that is done then I will just take a stab at it and see what you think.

lewisjkl:
I took a quick, perhaps somewhat naive, approach to creating the `partitionedStream`. Would love your feedback on this initial solution. I have a few specific questions:

1. I am not sure if there is a better way to build the partitions for the streams. My initial approach has been to require the user to provide a desired number of partitions. From there, map over the provided records to transform them into `CommittableConsumerRecord`s with offsets and partition numbers. Does this seem like a reasonable way to do this or do you have something else in mind?

2. The way that I split the input records into multiple substreams is using `Stream#broadcast`. The only downside to this approach is that it requires a `Concurrent[F]` to work. Is there another way to do this that wouldn't require an extra typeclass constraint? It seems a little unfortunate to need `Concurrent[F]` for a test implementation. Obviously I could perform this split using functions like `List#groupBy`, but I thought it would be nice to use the non-strictness of `Stream` here if possible.

3. I am starting to look at functions like `assignmentStream` and wonder what the expected behavior is there. Are we hoping that we will periodically and randomly perform a fake rebalance and emit elements from that stream? I think it would be really cool, but I also don't want to overdo it here.

Keep in mind that this is mainly just a quick pass at an implementation to start a discussion about possible solutions. And also there could be errors here because I still need to write tests 🙂

mwz:
> 1\. I am not sure if there is a better way to build the partitions for the streams. My initial approach has been to require the user to provide a desired number of partitions.

I guess you can infer the number of partitions from the supplied records, so there's probably no need to ask for it explicitly.

[...]

PR #318, opened 2020-01-28
first pass at creating mock partitionedStream

lewisjkl:
Here is a PR for issue #234. My initial approach has been to take the input `F[ConsumerRecord[K, V]]` along with a given number of partitions and use that to split the input into a stream of partitioned records. I am thinking this could be the easiest way to mimic the functionality that Kafka gives you out of the box.


aartigao:
I'm closing this draft because the preferred way is by using [fs2-kafka-mock](https://index.scala-lang.org/jchapuis/fs2-kafka-mock).

Thank you for your time 🙏🏽

PR #331, opened 2020-02-20
fix problem with emitting same data from stream after rebalance

vavr:
in some cases, streams on same partitions will never be closed.

Now, each stream has own unique internal ID, and if a request for a fetch for a partition with a large id arrives, then such a request is completed with empty chunk.

A couple of tests are written that fall in master, but pass with changes.


modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -184,20 +185,28 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
+                      fetch.completeRevoked(Chunk.empty) >>
+                        log(RevokedPreviousFetch(partition, streamId))
+                  )
+                  .void
               }
         }
 
     def completeRevoked =
-      deferred.complete((Chunk.empty, FetchCompletedReason.TopicPartitionRevoked))
+      deferred.complete((Chunk.empty, FetchCompletedReason.TopicPartitionRevoked)) >> ref.update {
```
vlovgr: Shouldn't this `ref.update` happen in `revoked` instead of here?

vavr: Oh yes, you are absolutely right!

vavr: I pushed the comment related changes

vavr:
@vlovgr what about this PR? This code is on our production about 2 weeks, and we have no problem with it!

vlovgr:
@vavr looks like the tests are still failing:
> Cause: java.lang.IllegalStateException: Attempting to complete a Deferred that has already been completed

vavr:
> @vavr looks like the tests are still failing:
> 
> > Cause: java.lang.IllegalStateException: Attempting to complete a Deferred that has already been completed

@vlovgr Can you please help with this fail? Locally it is not reproduced and I don't understand why is this test failed :(

We have this path on production about 5 months - and it works without any errors...

vlovgr:
@vavr I finally got around to looking at this properly. I've changed it to only track the most recent partition stream id, as there's no reason to keep historical ids. The changes looks good and tests are passing, so going to merge this. Thanks! 👍

PR #335, opened 2020-03-04
Add producer metrics

janstenpickle:
Similar to consumer metrics, expose underlying java client metrics in a tasteful way via the scala client.


janstenpickle:
@vlovgr sorry to be a pest, but is this PR ok?

vlovgr:
@janstenpickle Thanks for this! Unfortunately, this change is not binary compatible with v1.0.0. (It seems binary compatibility checking is disabled in `master`, fix is in #341.) This will either have to wait for v2.0.0 (not scheduled), or we have to add a new `abstract class KafkaProducer.Metrics` extending the existing `KafkaProducer`, and return that instead.

> [error] abstract method metrics()java.lang.Object in class fs2.kafka.KafkaProducer is present only in current version

janstenpickle:
Oh that's annoying, I suppose it would be wise to do the same for the consumer and deprecate the current `metrics` method.

janstenpickle:
@vlovgr I've moved the metrics method to the new abstract class as you suggested, sorry it took me so long!

Also I've created a similar abstract class for the consumer with no methods for consistency, however it feels a bit redundant. Would you prefer I only do this for the producer?

janstenpickle:
@vlovgr sorry to pester you, but it would be great to get this merged. Are the changes OK?

vlovgr:
@janstenpickle Sorry it's taken me a while to get around to this. Looks good, thanks. 👍

janstenpickle:
No worries, thanks @vlovgr!

PR #338, opened 2020-03-10
Provide explicit TopicPartition context in partitionedStream

jesperpi:
Proposed alternative to the work that DanXMoran has been working on https://github.com/fd4s/fs2-kafka/issues/132. This solution requires the client to generate the transactional producer settings themselves, but provides the tools to do so easily.

Could maybe add some sugar to create the producer staight out of the explicitPartitionedStream, but I would like to gather some feedback first.

This change adds a function to the `KafkaConsumerStream` which
explicitly returns the topic and partition along with the
partitioned stream of records.
This can be used in transactional producers to get a
`transactional.id` which is guaranteed to satisfy the scope
required by such an id.


jesperpi:
@vlovgr do you have any thoughts about this solution?

vlovgr:
@jesperpi Thanks for this! I like the idea, and it really seems quite simple.

Maybe @danxmoran has some thoughts on this?

danxmoran:
I agree, this looks good as is a lot simpler than what I'd experimented with doing!

jesperpi:
ok, do you think the interface can be improved? Also, should I write some tests of the producer?
This PR doesn't actually change the functionality; only the format returned by the function, so I'm scratching my head a bit about what testing would be valuable here.

jesperpi:
@vlovgr is there anything else needed before this is ready to be merged.
I notice the build is failing on forward source compatibility checks - I believe this is correct since I am adding a function to the KafkaConsumer.

jesperpi:
@vlovgr is this something you would consider merging?
As noted before, forward compatibility is broken by the new method, hence the build failure.
I don't see any natural way of avoiding this. Would be good to get your input.

LMnet:
The library could provide an even more informative method.

Kafka assignes partitions to the consumer using the next algorithm:
1. Revoke all previously assigned partitions.
2. Assign all new partitions at once.

See `ConsumerRebalanceListener` for more details.

So, the most basic version of `stream`  method could look like this:
```scala
def partitionsMapStream: Stream[F, Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]]]
```
Each element of the outer stream is a map with currently assigned partitions with its streams.

I plan to make pull request with this method in the near future.

jesperpi:
@LMnet That sounds like a great idea! Probably a significant improvement over what is proposed here. I look forward to seeing what you come up with :)

LMnet:
@jesperpi FYI: https://github.com/fd4s/fs2-kafka/pull/405

vlovgr:
Closing this in favour of #405.

PR #405, opened 2020-10-14
Partitions map stream

LMnet:
I added `partitionsMapStream: Stream[F, Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]]]` method to a `KafkaConsumer`.

This will be the most generic version across all stream methods on the `KafkaConsumer`. It reflects the Kafka rebalance process. With each rebalance Kafka revoke all partitions and after that assign all new partitions at once. In `partitionsMapStream` each element will be this assignment with the newly assigned partitions and it's corresponding streams. All other `stream` methods in my branch are based on `partitionsMapStream`.

Implementation was pretty simple because the old `partitionedStream` implementation contains all needed information. I just exposed all these details about partitions and assignments in the public API.

I believe, that `partitionsMapStream` would be useful in cases when a library user needs to do some custom stuff with each rebalance or for passing `TopicPartition` context to partition streams.

I'm using this method for a while in production for some custom rebalance handling.

Besides the new method, I refactored tests a bit. I notice some inconsistencies in KafkaConsumerSpec and eliminated them. More specifically:
* `KafkaConsumer#stream should correctly get partitions from topic via partitionsFor` renamed to a `KafkaConsumer#partitionsFor should correctly return partitions for topic` and moved from a `KafkaConsumer#stream` section to a `KafkaConsumer#partitionsFor` section.
* `KafkaConsumer#stream should fail when to get partitions from topic via partitionsFor with a small timeout` renamed to a `KafkaConsumer#partitionsFor should fail when timeout is too small` and moved from a `KafkaConsumer#stream` section to a `KafkaConsumer#partitionsFor` section.
* `KafkaConsumer#stream should get metrics` renamed to a `KafkaConsumer#metrics should return metrics` and moved from a `KafkaConsumer#stream` section to a `KafkaConsumer#metrics` section.
* `KafkaConsumer#stream should correctly unsubscribe` renamed to a `KafkaConsumer#unsubscribe should correctly unsubscribe` and moved from a `KafkaConsumer#stream` section to a `KafkaConsumer#unsubscribe` section.
* `KafkaConsumer#stream should handle rebalance` renamed to a `KafkaConsumer#partitionsMapStream should handle rebalance` and moved from a `KafkaConsumer#stream` section to a `KafkaConsumer#partitionsMapStream` section. Also, in this test, I added some new asserts because `partitionsMapStream` contains some more information about rebalance and it worth testing it.
* `KafkaConsumer#stream should close all streams on rebalance when using partitionedStream` renamed to a `KafkaConsumer#partitionsMapStream should close all old streams on rebalance` and moved from a `KafkaConsumer#stream` section to a `KafkaConsumer#partitionsMapStream` section. Also, this test was rewritten, because `partitionsMapStream` contains more information about rebalance and could provide some details for this test which was unavailable previously. Overall this test became a lot simpler.
* `KafkaConsumer#stream should close all streams on rebalance when using partitionedStream several times` test was deleted. I believe new `KafkaConsumer#partitionsMapStream should close all old streams on rebalance` is good enough to test both these cases.

Pull request contains 2 commits:
1. Some minor refactorings.
2. Main implementation.

Also, this pull request could be considered as an improved version of https://github.com/fd4s/fs2-kafka/pull/338.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -199,7 +199,8 @@ lazy val mimaSettings = Seq(
     Seq(
       ProblemFilters.exclude[Problem]("fs2.kafka.internal.*"),
       ProblemFilters.exclude[IncompatibleSignatureProblem]("*"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.kafka.KafkaProducer.resource")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.kafka.KafkaProducer.resource"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaConsumer.partitionsMapStream")
     )
     // format: on
   }
--- a/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -20,7 +20,9 @@ import fs2.kafka.internal.instances._
 import fs2.kafka.internal.KafkaConsumerActor._
 import fs2.kafka.internal.syntax._
 import java.util
+
 import org.apache.kafka.common.{Metric, MetricName, PartitionInfo, TopicPartition}
+
 import scala.collection.immutable.SortedSet
 import scala.concurrent.duration.FiniteDuration
 import scala.util.matching.Regex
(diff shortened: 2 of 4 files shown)


LMnet:
The build is read because of [this](https://github.com/fd4s/fs2-kafka/pull/406)

PR #406, opened 2020-10-15
Fix build

LMnet:
The latest update of scala 2.12.10 to 2.12.12 makes all build red because of this error:
```
[error] Test suite fs2.kafka.KafkaConsumerSpec failed with java.lang.NoClassDefFoundError: scala/math/Ordering$$anon$7.
[error] This may be due to the ClassLoaderLayeringStrategy (ScalaLibrary) used by your task.
[error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[error] The project class files are loaded in a separate class loader that is created for each test run.
[error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[error] With this approach, class loading may fail under the following conditions:
[error] 
[error]  * Dependencies use reflection to access classes in your project's classpath.
[error]    Java serialization/deserialization may cause this.
[error]  * An open package is accessed across layers. If the project's classes access or extend
[error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[error]    because the jvm enforces package private at the classloader level.
[error] 
[error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[error] in your configuration, for example:
[error] 
[error] set core / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set core / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
[error] 
[error] See ClassLoaderLayeringStrategy.scala for the full list of options.
```
I tried to investigate a bit and found [this](https://github.com/sbt/sbt/issues/4525) and [this](https://github.com/lucidsoftware/neo-sbt-scalafmt/issues/78). It looks like some sbt plugins manipulate scala version and cause this class loader issue. I'm not 100% sure that this is the root cause, but it looks pretty similar. 

I didn't find any valuable solution in a decent time, so I decided to just revert scala 2.12 update. Looks like this update is not necessary for the library and making all current builds green is very important for the pending pull requests.

[...]

PR #409, opened 2020-10-16
Added Traverse and Bitraverse instances to the CommittableConsumerRecord, CommittableProducerRecords, ConsumerRecord and ProducerRecord

LMnet:
In this pull request, I added `Traverse` and `Bitraverse` instances to the `CommittableConsumerRecord`, `CommittableProducerRecords`, `ConsumerRecord` and `ProducerRecord`. I was really missing these instances in everyday work.

Unfortunately, there are a few types that could benefit from these instances too, but they implementation does not give an opportunity to implement these instances. I'm talking about `ProducerResult`, `ProducerRecords`, and `TransactionalProducerRecords`. They all have a `P` (passthrough) type as the rightmost type. Scala's partial unification is right-biased. And because of this, we can only implement `Traverse` or `Bitraverse` for the rightmost types. Which would be inconsistent with all other types instances. I [created an issue](https://github.com/fd4s/fs2-kafka/issues/410) for moving `P` parameter to the left.


LMnet:
The build is red because of this: https://github.com/fd4s/fs2-kafka/pull/406

vlovgr:
Could you add the tests for checking the laws for these instances @LMnet?

LMnet:
> Could you add the tests for checking the laws for these instances @LMnet?

Ok, I will add them.

LMnet:
@vlovgr I added tests in the second commit.

For laws testing, I added cats-testkit-scalatest. And after that I found, that there are `CatsEquality.scala` and `CatsSuite.scala` files in the project. Looks like it was an old copy-paste before cats-testkit-scalatest was published as a separate artifact. So I deleted these copy-pasted files.

To properly test `Traverse` and `Bitraverse` instances I added `Eq` instances to a few types. All these new instances are covered with laws tests too.

Full list of changes of the second commit:
- Added Eq instances for:
  - Header
  - Headers
  - Timestamp
  - ConsumerRecord
  - ProducerRecord
  - CommittableOffset
  - CommittableConsumerRecord
  - CommittableProducerRecords
- Deleted CatsEquality.scala and CatsSuite.scala in favor of cats-testkit-scalatest.
- Added laws tests for all newly added Eq instances and
- Added laws tests for Traverse/Bitraverse instances for:
  - CommittableConsumerRecord
  - CommittableProducerRecords
  - ConsumerRecord
  - ProducerRecord

modules/core/src/main/scala/fs2/kafka/CommittableOffset.scala
```diff
@@ -131,4 +130,12 @@ object CommittableOffset {
...
   implicit def committableOffsetShow[F[_]]: Show[CommittableOffset[F]] =
     Show.fromToString
+
+  implicit def committableOffsetEq[F[_]]: Eq[CommittableOffset[F]] =
+    Eq.instance {
+      case (l, r) =>
+        l.topicPartition == r.topicPartition &&
+          l.offsetAndMetadata == r.offsetAndMetadata &&
+          l.consumerGroupId == r.consumerGroupId
+    }
```
LMnet, line 140: I'm not sure that this is 100% accurate because `CommittableOffset` also has a `commit: F[Unit]`. In my implementation, I just ignore this field.

modules/core/src/main/scala/fs2/kafka/Header.scala
```diff
@@ -62,4 +66,10 @@ object Header {
 
   implicit val headerShow: Show[Header] =
     Show.fromToString
+
+  implicit val headerEq: Eq[Header] =
+    Eq.instance {
+      case (l, r) =>
+        l.key === r.key && l.value.toVector === r.value.toVector
```
LMnet, line 73: Since there is no `Eq[Array[Byte]]` I converted arrays to a vector.

modules/core/src/test/scala/fs2/kafka/BaseGenerators.scala
```diff
@@ -48,6 +70,15 @@ trait BaseGenerators {
...
     Arbitrary(genCommittableOffset[F])
 
+  implicit def cogenCommittableOffset[F[_]]: Cogen[CommittableOffset[F]] =
+    Cogen { (seed: Seed, offset: CommittableOffset[F]) =>
+      (Cogen
+        .perturb(_: Seed, offset.topicPartition))
+        .andThen(Cogen.perturb(_, offset.offsetAndMetadata))
+        .andThen(Cogen.perturb(_, offset.consumerGroupId))
+        .apply(seed)
+    }
```
LMnet, line 80: Here I ignore `commit: F[Unit]` too, like in the `Eq` instance.

modules/core/src/main/scala/fs2/kafka/Header.scala
LMnet: I thought that converting to vector will have a too big performance impact. Now I'm using `sameElements`.

modules/core/src/main/scala/fs2/kafka/CommittableOffset.scala
bplommer: Yeah, that makes me slightly nervous. Do you have any thoughts on this @vlovgr?

vlovgr: I think this is the reason why these instances haven't been added before. In practice though, we should never have two consumers with the same (group-id, topic-partition, offset), so not sure how this would be an issue. 🤷‍♂️

LMnet: I think, in practice, this will not be a problem. 

But in terms of API, it feels a bit weird. `CommittableOffset` feels like pure data, but it contains `def commit: F[Unit]`, which is not data, but a behavior. Maybe it would be better if `CommittableOffset` will not have `commit` method? Maybe it would be more natural to have it on `KafkaConsumer` (like `def commit(committableOffset: CommitableOffset): F[Unit]`)?

Of course, this change will not be backward compatible, and I think it not worth it. Maybe, someday, in some major update :)

vlovgr: Would you mind opening an issue about this and we can schedule it for 2.0? :)

LMnet: Ok: https://github.com/fd4s/fs2-kafka/issues/462

bplommer (approved):
LGTM - amazing work!

bplommer:
Can we merge this @vlovgr?

bplommer:
Looks like we're now blocked on a cats-testkit-scalatest release for 3.0.0-M3. I've opened a PR at https://github.com/typelevel/cats-testkit-scalatest/pull/109

vlovgr (approved):

bplommer:
We might want to inline the dependency on `cats-testkit-scalatest` since it doesn't seem to be getting very prompt updates.

LMnet:
> We might want to inline the dependency on cats-testkit-scalatest since it doesn't seem to be getting very prompt updates.

It would be sad if you have to do it. Maybe we could find some of the maintainers in gitter?

bplommer:
> > We might want to inline the dependency on cats-testkit-scalatest since it doesn't seem to be getting very prompt updates.
> 
> 
> 
> It would be sad if you have to do it. Maybe we could find some of the maintainers in gitter?

I asked on `cats-dev` a few days ago but didn't get a reply. This will be a blocker every time there's a new Scala 3 milestone, so I'm reluctant to add a new dependency that  creates extra maintenance overhead unless there's a really strong reason.

LMnet:
Well, it's reasonable.

At least let's add a TODO comment on this inlined stuff to remove it in the future.

bplommer:
There's now a cat-testkit-scalatest release, so let's just use that.

There are some compile errors with Scala 3 - I can fix them later today.

bplommer:
Thanks @LMnet - sorry for taking so long over this!

PR #416, opened 2020-10-22
Scala 3 support

### LINKED ISSUE #415: Scala 3 support
Cats-effect and fs2 both support `0.27.0-RC1`, so I think we're unblocked for that.

bplommer:
Resolves #415


modules/core/src/test/scala/fs2/kafka/KafkaAdminClientSpec.scala
```diff
@@ -242,7 +242,7 @@ final class KafkaAdminClientSpec extends BaseKafkaSpec {
       }
     }
 
-    it("should support ACLs-related functionality") {
+    ignore("should support ACLs-related functionality") {
```
bplommer, line 245: On 3.0.0-M3 This was failing with an binary compatibility error in the Kafka broker - fairly sure it's to do with the way Kafka inlines the standard library as mentioned in https://github.com/embeddedkafka/embedded-kafka#important-known-limitation.

I think we should consider moving to containerised Kafka for tests.

LMnet: Migration to testcontainers-scala would be pretty simple.

bplommer: Yeah, just need to change how we provide custom broker config

vlovgr (approved):

PR #417, opened 2020-10-23
Port to cats-effect 3.0

bplommer:
Should be ready for a milestone release.


SystemFw:
cats-effect M2 is out, and it has Dispatcher

docs/src/main/mdoc/admin.md
```diff
@@ -119,7 +117,7 @@ There are functions available for working with consumer groups.
 ```scala mdoc:silent
 import cats.Parallel
 
-def consumerGroupOperations[F[_]: Concurrent: ContextShift: Parallel]: F[Unit] =
+def consumerGroupOperations[F[_]: Concurrent: Parallel]: F[Unit] =
```
SystemFw, line 120: only `Concurrent` is enough

bplommer:
but cats-effect M2 won't be binary compatible with fs2 M1, right?

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -331,73 +333,79 @@ private[kafka] object KafkaConsumer {
...
-            case ExitCase.Completed => polls.cancel
-            case _                  => F.unit
-          }, actor.cancel)
+  )(implicit F: Async[F]): F[KafkaConsumer[F, K, V]] = {
+     val actorFiber = F.uncancelable { _ =>
+          actor.join.flatMap { 
+            case Outcome.Succeeded(_) => polls.cancel
+            case _ => F.unit
+          }
+        }.onCancel(actor.cancel)
```
SystemFw, line 391: when will this trigger, given that you've made it uncancelable? 
You've actually made it uncancelable in two different ways: not using `poll`, and `start`ing

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -331,73 +333,79 @@ private[kafka] object KafkaConsumer {
...
-            case ExitCase.Completed => actor.cancel
-            case _                  => F.unit
-          }, polls.cancel)
+          F.uncancelable {_ => 
+            polls.join.flatMap {
+              case Outcome.Succeeded(_) => actor.cancel
+              case _ => F.unit
+            }
+          }.onCancel(polls.cancel)
+          .start
```
SystemFw, line 401: same

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -331,73 +333,79 @@ private[kafka] object KafkaConsumer {
...
 
-        actorFiber combine pollsFiber
-      }
+                  F.race(actorFiber, pollsFiber).map(_.merge).map{f =>
+    new KafkaConsumer[F, K, V] {
```
SystemFw, line 404: feels like this logic has changed, we probably want to revisit

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -416,32 +424,32 @@ private[kafka] object KafkaConsumer {
...
             dequeueDone <- Deferred[F, Unit]
             shutdown = F.race(fiber.join.attempt, dequeueDone.get).void
-            stopReqs <- Deferred.tryable[F, Unit]
+            stopReqs <- Deferred[F, Unit]
             _ <- partitions.enqueue1 {
               Stream.eval {
                 def fetchPartition(deferred: Deferred[F, PartitionRequest]): F[Unit] = {
                   val request = Request.Fetch(partition, streamId, partitionStreamId, deferred)
                   val fetch = requests.enqueue1(request) >> deferred.get
                   F.race(shutdown, fetch).flatMap {
```
SystemFw, line 433: Might want to introduce some syntax as part of this

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -416,32 +424,32 @@ private[kafka] object KafkaConsumer {
...
                     case Left(()) =>
-                      stopReqs.complete(())
+                      stopReqs.complete(()).void
 
                     case Right((chunk, reason)) =>
                       val enqueueChunk =
                         if (chunk.nonEmpty)
                           chunks.enqueue1(Some(chunk))
                         else F.unit
```
SystemFw, line 441: good opportunity for `whenA`

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -331,73 +333,79 @@ private[kafka] object KafkaConsumer {
...
     withConsumer: WithConsumer[F]
-  )(implicit F: Concurrent[F]): KafkaConsumer[F, K, V] =
-    new KafkaConsumer[F, K, V] {
-      override val fiber: Fiber[F, Unit] = {
-        val actorFiber =
-          Fiber[F, Unit](F.guaranteeCase(actor.join) {
-            case ExitCase.Completed => polls.cancel
-            case _                  => F.unit
-          }, actor.cancel)
+  )(implicit F: Async[F]): F[KafkaConsumer[F, K, V]] = {
```
SystemFw, line 385: I _think_ only `Concurrent` should be enough for this specific function (but I don't have a compiler, might be messing some constraint in the body)

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -534,7 +542,7 @@ private[kafka] object KafkaConsumer {
           streamId <- Stream.eval(streamIdRef.modify(n => (n + 1, n)))
           partitionStreamIdRef <- Stream.eval(Ref.of[F, Int](0))
           _ <- Stream.eval(initialEnqueue(streamId, partitions, partitionStreamIdRef))
```
SystemFw, line 544: Btw, it's better to merge all these `eval`s into one, since each `eval` has a `race`  in it for interruption

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -8,7 +8,7 @@ package fs2.kafka
 
 import cats.Apply
 import cats.effect._
-import cats.effect.concurrent.Deferred
+import cats.effect.kernel.Deferred
```
SystemFw, line 11: import from `core`, `kernel` imports ought to only be used when `core` is not a dependency

bplommer:
Though I suppose runtime linking errors aren't much of a concern until it starts compiling.

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -333,7 +339,7 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
       val commit =
         Deferred[F, Either[Throwable, Unit]].flatMap { deferred =>
           requests.enqueue1(Request.Commit(offsets, deferred)) >>
-            F.race(timer.sleep(settings.commitTimeout), deferred.get.rethrow)
+            F.race(F.sleep(settings.commitTimeout), deferred.get.rethrow)
```
SystemFw, line 342: why not `.timeout` here?

modules/core/src/main/scala/fs2/kafka/internal/WithConsumer.scala
```diff
@@ -19,37 +19,28 @@ private[kafka] object WithConsumer {
...
-                override def apply[A](f: KafkaByteConsumer => F[A]): F[A] =
-                  synchronizedConsumer.use { consumer =>
-                    context.blockOn(blocker) {
-                      f(consumer)
-                    }
-                  }
-              }
+    Resource[F, WithConsumer[F]] {
+      settings.createConsumer
+        .flatMap(Synchronized[F].of)
```
SystemFw, line 27: we might want to replace this with `Semaphore` now

modules/core/src/main/scala/fs2/kafka/internal/WithAdminClient.scala
```diff
@@ -20,34 +20,23 @@ private[kafka] sealed abstract class WithAdminClient[F[_]] {
...
         settings.createAdminClient.map { adminClient =>
           val withAdminClient =
             new WithAdminClient[F] {
               override def apply[A](f: AdminClient => KafkaFuture[A]): F[A] =
-                context.blockOn(blocker) {
-                  F.suspend(f(adminClient).cancelable)
-                }
+                ???
+                // context.blockOn(blocker) {
+                //   F.suspend(f(adminClient).cancelable)
```
SystemFw, line 31: mm interesting. Might need to use `blocking` first, and then convert the `Future`, but needs more thought

modules/core/src/main/scala/fs2/kafka/internal/WithConsumer.scala
```diff
@@ -19,37 +19,28 @@ private[kafka] object WithConsumer {
...
+        .map { synchronizedConsumer =>
+          val withConsumer =
+            new WithConsumer[F] {
+              override def apply[A](f: KafkaByteConsumer => F[A]): F[A] =
+                synchronizedConsumer.use { consumer =>
+                  F.defer(f(consumer))
```
SystemFw, line 33: needs `blocking` here, which you can't quite use, we need to think about this

modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
```diff
@@ -19,70 +19,54 @@ private[kafka] object WithProducer {
...
-      settings.blocker
-        .map(Resource.pure[F, Blocker])
-        .getOrElse(Blockers.producer)
+    Resource[F, WithProducer[F]] {
+      settings.createProducer.map { producer =>
+        val withProducer =
+          new WithProducer[F] {
+            override def apply[A](f: KafkaByteProducer => F[A]): F[A] =
+              F.defer {
+                f(producer)
```
SystemFw, line 30: same problem with `blocker`

SystemFw: Although maybe you are wrapping that somewhere else?

SystemFw:
> Though I suppose runtime linking errors aren't much of a concern until it starts compiling

yeah exactly :)

bplommer:
Thanks for the detailed feedback @SystemFw! To be honest I'm just playing compiler whack-a-mole so far but those comments will be super helpful later on 🙂

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
bplommer: do we actually need `core` as a dependency?

SystemFw: you already have it from fs2

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
bplommer: We also need `Sync` for suspending side-effects, so I went with `Async` as the least upper bound

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
bplommer: no idea. changed.

modules/core/src/main/scala/fs2/kafka/internal/WithAdminClient.scala
bplommer: Yeah, need to to grok what's going on with `cancelable`.

modules/core/src/main/scala/fs2/kafka/internal/WithConsumer.scala
bplommer: Oh, so in CE3 we don't have a way to shift an already-suspended effect to a blocking EC?

bplommer: It seems every invocation of this involves `F.delay` - if we change this to take a thunk we can use `blocking` and simplify the call site

modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
bplommer: I've changed from `delay` to `blocking` at the call site. (What I suggested above won't straightforwardly work where this is used by `TransactionalProducer`.

modules/core/src/main/scala/fs2/kafka/internal/WithAdminClient.scala
bplommer: Gave this a try with 50562c6 - I'm not at all sure whether it's correct though.

modules/core/src/main/scala/fs2/kafka/internal/WithConsumer.scala
SystemFw: > we don't have a way to shift an already-suspended effect to a blocking EC

we do (`evalOn`), but no direct access to _the_ blocking EC, for reasons

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -415,23 +415,21 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
-          }
+          ref
+            .updateAndGet(_.withRecords(newRecords))
+            .log(StoredRecords(newRecords, _))
+            .whenA(newRecords.nonEmpty)
         } else {
           val allRecords = state.records combine newRecords
 
-          if (allRecords.nonEmpty) {
+          F.whenA(allRecords.nonEmpty) {
```
SystemFw, line 409: there is syntax for this, `doSomething.whenA(thing)`. Probably want to check how @vlovgr feels about it, personally I find `F.thing { ... }` a bit hard to read

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -117,12 +119,11 @@ object KafkaProducer {
...
                   val complete =
                     deferred.complete {
-                      if (exception == null)
-                        Right((record, metadata))
+                      if (exception == null) Right((record, metadata))
                       else Left(exception)
                     }
 
-                  F.runAsync(complete)(_ => IO.unit).unsafeRunSync()
+                  dispatcher.unsafeRunSync(complete.void)
```
bplommer, line 126: I'm not sure what the reason for using `runAsync` was here, so not sure whether this will work

LMnet: That's because `F.async` will wait for the callback call and will not release `F` from the whole `commitAsync` method. But in the java client, a callback for the `commitAsync` could be resolved only with the `poll` calls. So, to make it work we need to release `commitAsync` right after the call and call `poll` after that. This is not how `F.async` works.

bplommer: I'll need to take a while to get my head around that, but thanks 🙂

SystemFw: I think Ben is asking about `runAsync` vs `unsafeRunSync` here, rather than `async` vs `delay`, but you make a good point regardless.
Is there a reason why this cannot be rewritten as
```
F.async_ { cb =>
  producer.send(
        javaRecord,
        callback { (metadata, exception) =>
              if (exception == null) cb(Right((record, metadata)))
              else cb(Left(exception))
         })
     }.start.map(_.join)
```

LMnet: I tried and it just didn't work for me. I didn't make a deep investigation though, so maybe I was doing something wrong. Theoretically, it should work. But callback just doesn't invoked for me when I'm trying it. Maybe because `poll` and this callback should be done in the same thread (that's how java client works)?

bplommer: @LMnet when you tried this did you do `.start.map(_.join)`? Because I previously tried changing the CE2 code to use `async` but without `start` and the tests froze like you're reporting, but now I'm trying it with `start` and it seems to work.

bplommer: If that does work, hopefully we can remove the `ConcurrentEffect` requirement with CE2.

SystemFw: `joinAndEmbedNever` is the CE3 equivalent

LMnet: Unfortunately, I don't remember exactly. I think I tried `.start.map(_.join)`, but I may be wrong.

Nevertheless, if it works — it's ok. However, to be completely sure, I would check that the thread which calls `callback` is the same thread, that calls `poll`.

SystemFw: >  I would check that the thread which calls callback is the same thread, that calls poll.

do you have any links to code or doc that specify this? I haven't seen it as a requirement for this method (although other parts of Kafka do require it)

bplommer: > I would check that the thread which calls `callback` is the same thread, that calls `poll`.

Which `poll` are we talking about? I can only find references to `poll` in the consumer, not the producer.

SystemFw: > I can only find references to poll in the consumer, not the producer.

that's my impression as well

LMnet: Oh, guys, I'm sorry, but I thought we are talking about the `KafkaConsumer` It has almost the same pattern with callback in the `commitAsync` method.

SystemFw: ah cool, worth keeping that comment in mind for the consumer anyway :)

bplommer: Does that mean we could make a similar change in the consumer if we run `poll` and the callback on a dedicated single-thread execution context?

LMnet: Specifically, I was talking about this: https://github.com/fd4s/fs2-kafka/pull/417/files?file-filters%5B%5D=.scala#diff-bcbdc586bf79fc42fcb6c0c4e87ec601c395715522f3af680cf988c36cc4d1eaR212

LMnet: > Does that mean we could make a similar change in the consumer if we run poll and the callback on a dedicated single-thread execution context?
It would be really cool, because atm there are a few `unsafeRunSync` calls in the consumer. I will be happy to get rid of them if possible.

bplommer: I've opened a PR that changes this on the producer side - https://github.com/fd4s/fs2-kafka/pull/418/files#diff-c4655864f9261fc7a30abdbdd8df44a5d4ed192dcdba30f5ab511b4f7dc8cb8f

bplommer:
This is nearly ready for a milestone release - just needs some cleaning up and getting a few of the tests to compile.

modules/core/src/main/scala/fs2/kafka/consumer/KafkaConsumerLifecycle.scala
```diff
@@ -30,5 +30,5 @@ trait KafkaConsumerLifecycle[F[_]] {
     * shut down. Most of the time, when you're only using the streams
     * provided by [[KafkaConsumer]], there is no need to use this.
     */
-  def fiber: Fiber[F, Unit]
+  def fiber: fs2.kafka.FakeFiber[F, Unit]
```
bplommer, line 33: todo: change this to something more reasonable

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -77,78 +76,123 @@ sealed abstract class KafkaConsumer[F[_], K, V]
...
-
-        val pollsFiber =
-          Fiber[F, Unit](polls.join.guaranteeCase {
-            case ExitCase.Completed => actor.cancel
-            case _                  => F.unit
-          }, polls.cancel)
-
-        actorFiber combine pollsFiber
-      }
+      def terminate: F[Unit] = actor.cancel *> polls.cancel
```
bplommer, line 180:
```suggestion
      def terminate: F[Unit] = fiber.join
```

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -77,78 +76,123 @@ sealed abstract class KafkaConsumer[F[_], K, V]
...
+          oc match {
+            case Outcome.Succeeded(_) => polls.cancel
+            case _                    => F.unit
+          }
+        } *> polls.join.guaranteeCase { oc: Outcome[F, Throwable, Unit] =>
+          oc match {
+            case Outcome.Succeeded(_) => actor.cancel
+            case _                    => F.unit
+          }
+        }
```
bplommer, line 193:
```suggestion
      def awaitTermination: F[Unit] = fiber.join
```

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
bplommer: > @LMnet I think Ben is asking about `runAsync` vs `unsafeRunSync` here, rather than `async` vs `delay`, but you make a good point regardless.
> Is there a reason why this cannot be rewritten as
> 
> ```
> F.async_ { cb =>
>   producer.send(
>       javaRecord,
>       callback { (metadata, exception) =>
>          if (exception == null) cb(Right((record, metadata)))
>          else cb(Left(exception))
>         }
>      )
>   }.start.map(_.join)
> ```

As it turns out, yes: https://github.com/fd4s/fs2-kafka/issues/477

bplommer:
I think we're ready to release a milestone once there's a new milestone for fs2-kafka. @SystemFw would you mind taking another look through this?

modules/core/src/main/scala/fs2/kafka/internal/FakeFiber.scala
```diff
@@ -0,0 +1,39 @@
...
+package fs2.kafka.internal
+
+import cats.effect.{Concurrent, Outcome}
+import cats.syntax.all._
+import cats.effect.syntax.all._
+
+/** A wrapper for `cancel` and `join` effects used to terminate and await
+  *  termination of running processes, ported from `Fiber` in cats-effect 2.
+  */
+case class FakeFiber[F[_]](join: F[Unit], cancel: F[Unit])(implicit F: Concurrent[F]) {
```
vlovgr, line 16: Should be `private[kafka]`?

modules/core/src/main/scala/fs2/kafka/internal/syntax.scala
```diff
@@ -194,22 +195,24 @@ private[kafka] object syntax {
     def void: KafkaFuture[Unit] =
       map(_ => ())
 
-    def cancelToken[F[_]](implicit F: Sync[F]): CancelToken[F] =
-      F.delay { future.cancel(true); () }
+    def cancelToken[F[_]](implicit F: Async[F]): F[Option[F[Unit]]] =
+      F.blocking { future.cancel(true); () }.start.as(None)
```
vlovgr, line 199: This looks strange with always returning `None`.

bplommer: The inner value is an optional cancellation token. You can't cancel the cancellation of the future, so that's why it's None - I think this is correct?

bplommer: Maybe the `as(None)` should move to the call site though

bplommer: Actually - we can return the cancel token for the cancelation fiber. updating.

SystemFw (approved):

PR #418, opened 2020-10-26
Weaken Producer requirements + refactoring

bplommer:
I'm opening this against the CE2 version to minimise the differences with CE3 in #417.

It does a few different things but they're organised into separate commits.


modules/core/src/main/scala/fs2/kafka/internal/WithConsumer.scala
```diff
@@ -7,12 +7,14 @@
...
 
 import cats.effect.{Blocker, Concurrent, ContextShift, Resource}
+import cats.effect.concurrent.Semaphore
 import cats.implicits._
 import fs2.kafka.{KafkaByteConsumer, ConsumerSettings}
 import fs2.kafka.internal.syntax._
 
 private[kafka] sealed abstract class WithConsumer[F[_]] {
   def apply[A](f: KafkaByteConsumer => F[A]): F[A]
+  def blocking[A](f: KafkaByteConsumer => A): F[A]
```
vlovgr, line 17: I'm slightly 👎 against `blocking`, since it's somewhat easy to return `F[A]` here and discard it by accident. Although if we still want it, I'd suggest we add it to `WithProducer` as well to keep them consistent.

SystemFw: This is mandated by the design of cats-effect pretty much. Without it, you will have to manually thread an `EC` for blocking op, whereas `Blocker` disappears with this choice

bplommer: You *could* do `Sync[F].blocking` at the use site, no?

SystemFw: oh yeah sure

bplommer: I still prefer `blocking` though. I think the errors Viktor describes will be much easier to catch (either with the compiler or tests) than an error with using `delay` instead of `blocking` in calling code. (And it's less boilerplatey too)

vlovgr: I _think_ all calls are blocking, so no need to differentiate between the two. With CE3 we would just merge the two into the signature of `blocking`, so maybe do that here instead then (and keep it as `apply`)?

bplommer: You're right we should do the same thing with `WithProducer` - I'll make that change. There's also uses in `KafkaConsumerActor` I still need to change.

Unfortunately I don't think we can completely get rid of the current `apply` in `WithProducer`, because of the way [TransactionalKafkaProducer uses `bracket`](https://github.com/fd4s/fs2-kafka/blob/114136ad0fe9063fd799d75c3c938d78f63013ec/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala#L70-L91). Maybe we should rename `apply` to `defer` (the name for `suspend` in CE3) though.

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -106,28 +106,25 @@ object KafkaProducer {
...
+                  else Left(exception)
                 }
-              )
-            }
-            .as(deferred.get.rethrow)
-        }
+              }
+            )
+            ()
+          }
```
SystemFw, line 125: this is CE2, so you need a `guarantee(context.shift)` here

SystemFw: although maybe it doesn't matter given that you `start`. Might want to test it if you can be bothered

SystemFw (approved):
Pending Viktor's comment on `blocking` ofc

vlovgr (approved):

bplommer:
@SystemFw Would you mind taking a look at the new changes? I'm not terribly confident about whether things are running on the right EC.

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
bplommer: I've put it in the wrong place now, right?

SystemFw:
Yeah I'll have a look tomorrow

bplommer:
@SystemFw do you think you'll be able to get to this soon, or would it be better if I separate out the more straightforward changes into a separate PR that I can go ahead and merge so as to reduce the size of this one?

SystemFw:
sorry, forgot about this. Are there specific things you want me to check, or just a general review?

bplommer:
No worries - I'll go through the changes and tag the things I'd like you to check.

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -82,17 +82,15 @@ object KafkaProducer {
             override def produce[P](
               records: ProducerRecords[K, V, P]
             ): F[F[ProducerResult[K, V, P]]] = {
-              withProducer { producer =>
+              withProducer { (producer, _) =>
                 records.records
                   .traverse(produceRecord(keySerializer, valueSerializer, producer))
                   .map(_.sequence.map(ProducerResult(_, records.passthrough)))
               }
             }
```
bplommer, line 90: With this change, this no longer shifts to the blocking pool - am I right in changing this, since it's invoking `produceRecord` which is async? (@SystemFw )

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -106,28 +104,27 @@ object KafkaProducer {
...
                 }
-              )
-            }
-            .as(deferred.get.rethrow)
-        }
+              }
+            )
+            ()
+          }
+          .guarantee(context.shift)
```
bplommer, line 125: Is this in the right place? (Don't think I can be bothered to master these CE2 nuances at this stage)(@SystemFw )

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -169,69 +160,52 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
           )
-        }.attempt
-      }
-
-    commit.flatMap {
-      case Right(()) => F.unit
-      case Left(e)   => request.deferred.complete(Left(e))
+        }
+        .handleErrorWith(e => F.delay(request.callback(Left(e))))
     }
```
bplommer, line 206: As above - am I correct in changing this to not shift anything to the blocking pool? (and generally, could you check the way I've replaced the `Deferred` + `runAsync` with a side-effecting callback?)(@SystemFw )

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -364,20 +337,16 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
+        F.asyncF[Unit] { cb =>
+            requests.enqueue1(Request.Commit(offsets, cb))
+          }
+          .guarantee(context.shift)
+          .timeoutTo(settings.commitTimeout, F.raiseError[Unit] {
+            CommitTimeoutException(
+              settings.commitTimeout,
+              offsets
+            )
+          })
```
bplommer, line 350: Is this correct? Is the `.guarantee(context.shift)` needed, and if so is it in the right place? (@SystemFw )

bplommer:
I think I've tagged you now in all the places where I wanted a third pair of eyes.

SystemFw:
Cool, let me have a look

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
SystemFw: yeah this is correct

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
SystemFw: yeah, that's right

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
SystemFw: ah, that's interesting because you have a "split" async basically. You can put it here, or after the call to `commitAsync` (assuming that's the only place the callback is used non trivially)

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
SystemFw: This makes sense to me, see comment below about `shift`. Also in CE3 we might want to do this differently, but it's too hard to think about both at this stage, let's get this going instead

bplommer: Cool, so are you happy for me to go ahead and merge this?

SystemFw: yeah I would

PR #420, opened 2020-10-28
Added commitSync and commitAsync methods to the KafkaConsumer

LMnet:
I added `commitSync` and` commitAsync` methods. I decided to add both methods (not just `commitAsync`) because they have slightly different semantics. For example, `commitAsync` is based on` poll` calls, but `commitSync` not.

I also added tests for these methods. I managed to reuse the existed test code for this.


LMnet:
I rebased a branch to the current master and implementation is slightly changed:
1. I added the `KafkaCommit` trait with the new methods.
2. New actor requests are now named `ManualCommitAsync` and `ManualCommitSync`. Now I can't reuse `Commit` for async commit, because it doesn't have a suitable callback.
3. I refactored a bit, to reuse existing parts of the code.

bplommer (approved):

vlovgr (approved):

bplommer:
Thanks!

PR #421, opened 2020-10-29
Added stopConsuming method to the KafkaConsumer

LMnet:
I added the `stopConsuming` method to the `KafkaConsumer`. This method could be used to implement a graceful shutdown.

This method has a few effects:
1. After its call no more data will be fetched from Kafka through the `poll` method.
2. All currently running streams will continue to run until all in-flight messages will be processed.
   It means that streams will be completed when all fetched messages will be processed.

More details about its usage I added in the docs. Also, I added tests and some examples could be found there.

About design decisions. This method is pretty low-level and not opinionated. It's a building block for making your own graceful shutdown, not a ready-made solution. At first, I tried to add some high-level graceful shutdown functionality. But I found that with the current library design it would be problematic in terms of API. All my experiments end with pretty complex and at the same time not very extensible implementations. So, I decided to provide only a low-level API with detailed documentation.

About implementation. It turned out that it could be pretty simple. We just need to stop fetching new records from the `KafkaConsumerActor` and stop all running streams after that. To stop streams I made them NoneTerminatedQueue and just publish `None` to them on `stopConsuming` call.


bplommer (approved):

LMnet:
I updated the branch to the current master. I had to make only one additional change after updating: I move `stopConsuming` to the `KafkaConsume` trait.

modules/core/src/main/scala/fs2/kafka/consumer/KafkaConsume.scala
```diff
@@ -66,4 +66,20 @@ trait KafkaConsume[F[_], K, V] {
...
+    * 1. After this call no more data will be fetched from Kafka through the `poll` method.
+    * 2. All currently running streams will continue to run until all in-flight messages will be processed.
+    *    It means that streams will be completed when all fetched messages will be processed.<br>
+    * <br>
+    * If some of the [[stream]] methods will be called after [[stopConsuming]] call,
+    * these methods will return empty streams.<br>
+    * <br>
+    * More than one call of [[stopConsuming]] will have no effect.
+    */
+  def stopConsuming: F[Unit]
```
bplommer, line 84: I think I'd tend to put this in the `Lifecycle` trait rather than the `Consume` trait, but not sure - any thoughts? (cc @vlovgr)

I'm happy to make and push that change if you (@LMnet) are happy with that, since it's my fault this PR needed updating 🙂

LMnet: I hesitated between `KafkaConsume` and `KafkaLifecycle` and decided that `KafkaConsume` is better for this method because it's highly coupled with the stream methods.

vlovgr: I'm also leaning slightly towards `KafkaConsume`, but I'm ultimately happy either way. :)

LMnet:
Should I do something else to make it merged?

vlovgr (approved):

PR #429, opened 2020-11-12
Decompose consumer into traits

### LINKED ISSUE #380: KafkaConsumer should not be sealed
`KafkaConsumer` is a sealed abstract class. That serves no purpose since it is not an enumeration that gets pattern matched somewhere and it has no concrete implementations.

On the flip side, being sealed means you can't mock it easily for testing. `KafkaProducer` is also abstract but not sealed.

bplommer:
Resolves #380


bplommer:
This is failing MiMa and I have no idea why:
```
[error] fs2-kafka: Failed binary compatibility check against com.github.fd4s:fs2-kafka_2.12:1.1.0! Found 1 potential problems (filtered 130)
[error]  * abstract method partitionsMapStream()fs2.internal.FreeC in interface fs2.kafka.consumer.Consume is inherited by class KafkaConsumer in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.partitionsMapStream")
```

bplommer:
Oh - looks like it's a false positive because `partitionsMapStream` was added since 1.1.0 and MiMa isn't taking into account that `KafkaConsumer` is sealed.

bplommer:
Finally passed CI 😅 Any thoughts on this @vlovgr ?

vlovgr:
Nice! Looks good. Have a couple of comments.

- Thinking if we should add a prefix to these traits rather than have them in a subpackage.

  Partly because:
  ```scala
  def test[F[_]: Assignment: Consume]
  ```
  is not as descriptive as:
  ```scala
  def test[F[_]: KafkaAssignment: KafkaConsume]
  ```
  or maybe there's a better prefix/naming here.

- We have #420 and #421 open, should we add `KafkaCommit` for #420 and put #421 together with `ConsumerFiber` in a renamed trait? Or do you have any other ideas?

bplommer:
> * Thinking if we should add a prefix to these traits rather than have them in a subpackage.
>   Partly because:
>   ```scala
>   def test[F[_]: Assignment: Consume]
>   ```
>   
>   
>   is not as descriptive as:
>   ```scala
>   def test[F[_]: KafkaAssignment: KafkaConsume]
>   ```
>   
>   
>   or maybe there's a better prefix/naming here.

The thing I like about the current approach is that you can choose how much of the package structure to import and how much to name at the use site, and get more or less explicit context accordingly - so you can have e.g.
```scala
import fs2.kafka
def test[F[_]: kafka.consumer.Assignment: kafka.consumer.Consume]
```
but you can also do `import fs2.kafka.consumer._` to get the more concise naming. 

However, maybe for consistency and clarity we should name all these traits after the capabilities they provide - so e.g. `ManageAssignments` instead of `Assignment`.

That said, I don't feel very strongly about this - what do you think?

> * We have #420 and #421 open, should we add `KafkaCommit` for #420 and put #421 together with `ConsumerFiber` in a renamed trait? Or do you have any other ideas?

Sounds good. We could rename `ConsumerFiber` to something like `ManageConsumerLifecycle`. I'd been going to suggest deprecating `fiber` and replacing it with specific methods to terminate / await termination of the consumer (especially in view of differences between CE2/CE3 fibers), so changing the name might be good from that point of view too.

vlovgr:
My personal take is that few people will actually do `import fs2.kafka` and then specify `kafka.consumer.Assignment` (mainly because of IDE auto-completion and auto-import) -- but I could be in the wrong here. But for that reason, I'm mainly concerned with the trait itself having a name which makes it somewhat clear we're talking about a Kafka (consumer) capability. 

> Sounds good. We could rename `ConsumerFiber` to something like `ManageConsumerLifecycle`. I'd been going to suggest deprecating `fiber` and replacing it with specific methods to terminate / await termination of the consumer (especially in view of differences between CE2/CE3 fibers), so changing the name might be good from that point of view too.

Agreed. The current `fiber` isn't really a `Fiber` in the proper sense anyway, so dedicated methods make more sense.

bplommer:
That's reasonable. I'll add a `Kafka` prefix

bplommer:
I'd be in favour of keeping the sub-package too though, for the sake of keeping the files manageable - what do you think?

vlovgr:
> I'd be in favour of keeping the sub-package too though, for the sake of keeping the files manageable - what do you think?

Sounds good to me. 👍

bplommer:
@vlovgr how does this look to you? Shall we merge this, then update #420 and #421 accordingly and merge them before the next release?

vlovgr (approved):
> @vlovgr how does this look to you? Shall we merge this, then update #420 and #421 accordingly and merge them before the next release?

Looks great and yes, that would indeed be good.

PR #436, opened 2020-11-27
Allow multiple KafkaProducer instances to share underlying resources

bplommer:
Adds a `KafkaProducer.Connection` class that represents a connection to the Kafka cluster and allows instantiating `KafkaProducer` without allocating new resources.

This really makes it clear that `ProducerSettings` represents two different things - settings for the Kafka producer itself (which is serialization-agnostic), and settings for serialization. I'm starting to think we should take the opportunity to make some breaking changes, including separating out those two groups of settings, and release a `2.0` version - then we can release the CE3 port as `3.0`. Thoughts @vlovgr ?


vlovgr (commented):
Looks good, but would suggest naming it `KafkaProducerConnection` and move to a separate file.

bplommer:
> Looks good, but would suggest naming it `KafkaProducerConnection` and move to a separate file.

sounds good

vlovgr (approved):

bplommer:
Will add tests before merging

PR #437, opened 2020-11-27
Improve ergonomics of instantiating producers

bplommer:
I think methods on companion objects are more idiomatic in Scala - they're also more discoverable, especially with autocompletion tooling. I'd be inclined to deprecate the package object methods and remove them in v2 - thoughts?


vlovgr (approved):
👍

PR #451, opened 2020-12-15
Update producer docs and code examples for version 1.1.0

marko-asplund:


bplommer (dismissed):
The problem here isn't that the docs are out of date - rather, they've been updated for changes in #437 that haven't yet been released. @vlovgr is there a way we can version the docs so what people see by default reflects the latest released version?

vlovgr:
> is there a way we can version the docs so what people see by default reflects the latest released version?

The easiest way is probably to keep a separate branch for the next release and only merge at time of release.

bplommer:
> > is there a way we can version the docs so what people see by default reflects the latest released version?
> 
> The easiest way is probably to keep a separate branch for the next release and only merge at time of release.

Sounds good - let's do that after releasing 1.2.0?

vlovgr:
> Sounds good - let's do that after releasing 1.2.0?

👍

PR #466, opened 2021-01-14
Move `P` type param in ProducerRecords to the left

### LINKED ISSUE #410: Move P type in ProducerResult, ProducerRecords and TransactionalProducerRecords to the left
In [this](https://github.com/fd4s/fs2-kafka/pull/409) pull request some of the library types got `Traverse` and `Bitraverse` instances. This improves their usability. 

Unfortunately, there are a few types that could benefit from these instances too, but they implementation does not give an opportunity to implement these instances. I'm talking about `ProducerResult`, `ProducerRecords`, and `TransactionalProducerRecords`. They all have a `P` (passthrough) type as the rightmost type. Scala's partial unification is right-biased. And because of this, we can only implement `Traverse` or `Bitraverse` for the rightmost types. Which would be inconsistent with all other types instances.

Because of this, I want to move `P` type to the left and after that implement `Traverse` and `Bitraverse` instances for `ProducerResult`, `ProducerRecords`, and `TransactionalProducerRecords`. This change would be not source compatible with the current codebase, so it could be done only in 2.0.0 version.

bplommer:
Resolves #410

I've created a `series/2.x` draft to open breaking PRs against.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
@@ -50,8 +50,8 @@ abstract class KafkaProducer[F[_], K, V] {
     *   but losing the order of produced records.
     */
   def produce[P](
-    records: ProducerRecords[K, V, P]
-  ): F[F[ProducerResult[K, V, P]]]
+    records: ProducerRecords[P, K, V]
+  ): F[F[ProducerResult[P, K, V]]]
 }
 
 object KafkaProducer {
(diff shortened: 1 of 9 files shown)


LMnet:
Nice! 

But there is also `TransactionalProducerRecords[F, K, V, P]` which should be `TransactionalProducerRecords[F, P, K, V]`.

bplommer:
> Nice!
> 
> But there is also `TransactionalProducerRecords[F, K, V, P]` which should be `TransactionalProducerRecords[F, P, K, V]`.

Good catch :+1:

vlovgr (approved):

PR #468, opened 2021-01-15
use github matrix for CI

bplommer:


.github/workflows/ci.yml
```diff
@@ -2,16 +2,18 @@ name: ci
 on:
   push:
     branches: ['*']
-  pull_request:
```
vlovgr, line 5: Deleting this means that we no longer build pull requests (coming from a fork) raised against the repository.

.github/workflows/ci.yml
```diff
@@ -2,16 +2,18 @@ name: ci
...
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
     - uses: olafurpg/setup-scala@v10
     - uses: coursier/cache-action@v5
-    - run: sbt validate
-    - uses: codecov/codecov-action@v1
+    - run: sbt ++${{ matrix.scala }} validate
+    - if: matrix.scala == '2.13.3'
+      run: sbt "++${{ matrix.scala }} docs/run"
```
vlovgr, line 19: I'm not a fan of having scala versions here, and essentially overriding whatever is defined in `build.sbt`.

build.sbt
```diff
@@ -310,15 +310,12 @@ def addCommandsAlias(name: String, values: List[String]) =
 addCommandsAlias(
   "validate",
   List(
```
vlovgr, line 312: Sadly, these changes mean you can no longer just run `sbt validate` (you have to run `sbt +validate`) locally to check everything passes.

vlovgr:
@bplommer I think it would be good to have parallel tests for time-saving reasons. But would like to investigate using something like [sbt-github-actions](https://github.com/djspiewak/sbt-github-actions) to avoid the issues in my comments above. In the meantime I'm reverting this with #471.

bplommer:
> @bplommer I think it would be good to have parallel tests for time-saving reasons. But would like to investigate using something like [sbt-github-actions](https://github.com/djspiewak/sbt-github-actions) to avoid the issues in my comments above. In the meantime I'm reverting this with #471.

Sounds reasonable - I'll look into using sbt-github-actions.

Sorry, should have left this open for review.

PR #469, opened 2021-01-15
replace embedded kafka with testcontainers

bplommer:
Needed because of binary compatibility issues with scala 3


bplommer:
@vlovgr this is now ready to review for real 🙂

license.txt
```diff
@@ -174,3 +174,30 @@
       of your accepting any such warranty or additional liability.
 
    END OF TERMS AND CONDITIONS
+
+-----
+
+This project (specifically BaseKafkaSpec.scala) contains code derived from
```
vlovgr, line 180: Could we add the license to `BaseKafkaSpec.scala` as well?

vlovgr (approved):

license.txt
bplommer: sure :+1:

bplommer: Done - by the way, should we add copyright/license headers to the test files?

PR #472, opened 2021-01-17
Use sbt-github-actions for ci, remove coverage

bplommer:


bplommer:
If the CI part works I'd suggest we do a fake release to validate publishing.

.github/workflows/ci.yml
```diff
@@ -1,17 +1,98 @@
...
+            ~/AppData/Local/Coursier/Cache/v1
+            ~/Library/Caches/Coursier/v1
+          key: ${{ runner.os }}-sbt-cache-v2-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
+
+      - env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+        run: sbt ++${{ matrix.scala }} ci-release docs/docusaurusPublishGhpages
```
vlovgr, line 98: I've never tried overriding the scala version for these commands, but let's try it out.

vlovgr (approved):
Let's remove the [Codecov badge](https://github.com/fd4s/fs2-kafka/blob/6f9df5d2c2eee196084f2a4d02d0c2ec71cff45c/website/pages/en/index.js#L98) as well while we're at it.

.github/workflows/ci.yml
bplommer: the matrix for release only contains the default scala version anyway, so I don't think it will change anything.

PR #476, opened 2021-01-17
remove deprecated aliases in 2.x branch

bplommer:


build.sbt
```diff
@@ -306,7 +300,7 @@ addCommandsAlias(
     "+coverage",
     "+test",
     "+coverageReport",
-    "+mimaReportBinaryIssues",
+    //"+mimaReportBinaryIssues",
```
vlovgr, line 303: I'd suggest to make a commit to the series-branch where we set `mimaPreviousArtificats` to empty instead.

vlovgr (approved):

PR #478, opened 2021-01-18
Fix producer regression in v1.2.0

### LINKED ISSUE #477: 1.2.0 regression: lost guarantee of order for producer
It looks like [this](https://github.com/fd4s/fs2-kafka/pull/418/files#diff-c4655864f9261fc7a30abdbdd8df44a5d4ed192dcdba30f5ab511b4f7dc8cb8fR112-R124) pull request slightly change the semantic of producing in fs2-kafka. Java Kafka producer gives a guarantee, that messages will be produced in the same order as `producer.send` calls. In the version of fs2-kafka before 1.2.0, this guarantee was maintained. But after 1.2.0 some of my tests, which rely on this guarantee fail.

I suppose it's somehow related to the `F.async`.

bplommer:
Resolves #477

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -189,7 +189,9 @@ lazy val publishSettings =
 lazy val mimaSettings = Seq(
   mimaPreviousArtifacts := {
     if (publishArtifact.value) {
-      Set(organization.value %% moduleName.value % (previousStableVersion in ThisBuild).value.get)
+      // Temporary change because v1.2.0 is a cursed release
+      // Set(organization.value %% moduleName.value % (previousStableVersion in ThisBuild).value.get)
+      Set(organization.value %% moduleName.value % "1.1.0")
     } else Set()
   },
   mimaBinaryIssueFilters ++= {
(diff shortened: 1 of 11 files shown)


bplommer:
@vlovgr do you think you'll be able to take a look at this this morning?

vlovgr (approved):
👍

vlovgr:
This is the v1.1.0 version:
```
    record =>
        Deferred[F, Either[Throwable, (ProducerRecord[K, V], RecordMetadata)]].flatMap { deferred =>
          F.delay {
              producer.send(
                javaRecord,
                callback { (metadata, exception) =>
                  val complete =
                    deferred.complete {
                      if (exception == null)
                        Right((record, metadata))
                      else Left(exception)
                    }

                  F.runAsync(complete)(_ => IO.unit).unsafeRunSync()
                }
              )
            }
            .as(deferred.get.rethrow)
        }
      }
```

Should we change to:

```

bplommer:
> Should we change to?:
> 
> ```
> F.runAsync(complete)(_ => IO.unit).unsafeRunSync()
> ```
> 
> Either way, `toIO` won't be an option for ce3, so might as well change it now. @bplommer

Is there a particular reason for `runAsync`? I think in CE3 we'll use [Dispatcher#unsafeRunSync](https://github.com/typelevel/cats-effect/blob/6c96233c42b7bbc4793b4b4c0a28057587bc6f1e/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala#L29) which is closer to what's in this PR than to v1.1.0

vlovgr:
I guess mainly to avoid the conversion to `IO`, and avoiding blocking (although there shouldn't be much contention on the `Deferred`).

bplommer:
> I guess mainly to avoid the conversion to `IO`, and avoiding blocking (although there shouldn't be much contention on the `Deferred`).

Oh ok - unfortunately I've already kicked off the v1.3.0 release, so shall we make that change in v1.3.1 along with #416?

vlovgr:
> Oh ok - unfortunately I've already kicked off the v1.3.0 release, so shall we make that change in v1.3.1 along with #416?

Sounds good. 👍

PR #497, opened 2021-01-22
Rearrange directory structure

bplommer:
This organizes files into `common`, `consumer`, `producer`, `producer/transactional` and `adminclient` directories, with `settings` subdirectories as appropriate.

This is in preparation for changing the package structure in 2.x, and potentially modularising the library in the future.


bplommer:
Converting this to a draft as it needs some more work to remove the dependence of `producer/transactional` on `consumer`

vlovgr:
How about naming `adminclient` just `admin`? (Same way `consumer` isn't named `consumerclient`.)

bplommer:
> How about naming `adminclient` just `admin`? (Same way `consumer` isn't named `consumerclient`.)

Sure.

I'm going to close this and open another PR that organizes things into modules, as I think that's the best way to make sure we get the dependencies right.

PR #499, opened 2021-01-22
add ScalaFix rules for deprecated package object methods

### LINKED ISSUE #491: ScalaFix migration rules
As we make more API changes and deprecations, having ScalaFix rules would be increasingly useful, especially as scala-steward can apply them automatically.

bplommer:
Unfortunately this isn't integrated into the main build as ScalaFix isn't available for Scala 3, but the rules can be run directly from GitHub as outlined [here](https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#using-github).

It also caught a deprecated usage in `package.scala` that I missed 😅

Resolves #491


vlovgr (approved):
Fantastic stuff! 👍

PR #502, opened 2021-01-25
Break build up into more modules

bplommer:


bplommer:
The change of module structure is causing this to spuriously fail the MiMa check. What do you think about doing a release with Mima disabled @vlovgr?

bplommer:
Any thoughts on this @vlovgr ? I'd at least like to change the package locations of the `JavaX` aliases before releasing v1.4.0 so we don't need to support the ones we currently have (that haven't been released).

bplommer:
Closing this for now - will revisit before 3.0 release

Issue #490, opened 2021-01-19
Remove `KafkaConsumer#fiber`


bplommer:
`KafkaConsumer#fiber` isn't a real cats-effect fiber, and it's not clear how to implement it in CE3.

Instead, provide `terminate` and `awaitTermination` methods to replace `fiber.cancel` and `fiber.join`, and deprecate `fiber` in 1.x

agustafson:
Will take a look at a fix for this now

PR #504, opened 2021-01-26
Remove `KafkaConsumer.fiber`

agustafson:
Attempt to fix https://github.com/fd4s/fs2-kafka/issues/490


agustafson:
Getting mima compatibility errors because we've added new methods, but not sure how we deal with them

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -78,20 +78,10 @@ sealed abstract class KafkaConsumer[F[_], K, V]
...
 object KafkaConsumer {
-  private def startConsumerActor[F[_], K, V](
-    requests: Queue[F, Request[F, K, V]],
-    polls: Queue[F, Request[F, K, V]],
-    actor: KafkaConsumerActor[F, K, V]
-  )(
-    implicit F: Concurrent[F],
-    context: ContextShift[F]
-  ): Resource[F, Fiber[F, Unit]] =
+  private def spawn[F[_]: Concurrent, A](fa: F[A]): Resource[F, Fiber[F, Unit]] =
```
bplommer, line 81: nice 👍

bplommer:
> Getting mima compatibility errors because we've added new methods, but not sure how we deal with them

Yeah, that's a pain. My best suggestion is to add the new methods to a `trait KafkaConsumerLifecycleV2[F[_]] extends KafkaConsumerLifecycle[F]` and replace `KafkaConsumerLifecycle` with `KafkaConsumerLifecycleV2` in the definition of `KafkaConsumer`.

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -554,6 +549,16 @@ object KafkaConsumer {
 
       override def toString: String =
         "KafkaConsumer$" + id
+
+      // Stop polling, then stop consuming, then stop handling messages
+      private def shutdown(graceful: Boolean): F[Unit] = {
+        val shutdownOp: Fiber[F, Unit] => F[Unit] = if (graceful) _.join else _.cancel
+        shutdownOp(polls).attempt.void *> stopConsuming *> shutdownOp(actor).attempt.void
+      }
```
bplommer, line 557: This looks nice to have but probably needs more tests. In the meantime, I think we should just have `terminate` do the same thing `fiber.cancel` and `awaitTermination` do the same thing as `fiber.join`, and then think about building higher-level functionality on top of that.

bplommer: We'd also want to update the [graceful shutdown](https://github.com/fd4s/fs2-kafka/blob/series/1.x/docs/src/main/mdoc/consumers.md#graceful-shutdown) section in the docs.

agustafson:
> > Getting mima compatibility errors because we've added new methods, but not sure how we deal with them
> 
> Yeah, that's a pain. My best suggestion is to add the new methods to a `trait KafkaConsumerLifecycleV2[F[_]] extends KafkaConsumerLifecycle[F]` and replace `KafkaConsumerLifecycle` with `KafkaConsumerLifecycleV2` in the definition of `KafkaConsumer`.

I still get
```
[error] fs2-kafka: Failed binary compatibility check against com.github.fd4s:fs2-kafka_2.13:1.3.1! Found 2 potential problems
[error]  * abstract method terminate()java.lang.Object in interface fs2.kafka.consumer.KafkaConsumerTerminatingLifecycle is inherited by class KafkaConsumer in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.terminate")
[error]  * abstract method awaitTermination()java.lang.Object in interface fs2.kafka.consumer.KafkaConsumerTerminatingLifecycle is inherited by class KafkaConsumer in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.awaitTermination")
```
because `KafkaConsumer` still exposes those methods, unless I've misunderstood what you were suggesting @bplommer. I moved the new `terminate` methods to `KafkaConsumerTerminatingLifecycle`, made it extend `KafkaConsumerLifecycle` and made `KafkaConsumer` extend `KafkaConsumerTerminatingLifecycle`

agustafson:
> > > Getting mima compatibility errors because we've added new methods, but not sure how we deal with them
> > 
> > 
> > Yeah, that's a pain. My best suggestion is to add the new methods to a `trait KafkaConsumerLifecycleV2[F[_]] extends KafkaConsumerLifecycle[F]` and replace `KafkaConsumerLifecycle` with `KafkaConsumerLifecycleV2` in the definition of `KafkaConsumer`.
> 
> I still get
> 
> ```
> [error] fs2-kafka: Failed binary compatibility check against com.github.fd4s:fs2-kafka_2.13:1.3.1! Found 2 potential problems
> [error]  * abstract method terminate()java.lang.Object in interface fs2.kafka.consumer.KafkaConsumerTerminatingLifecycle is inherited by class KafkaConsumer in current version.
> [error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.terminate")
> [error]  * abstract method awaitTermination()java.lang.Object in interface fs2.kafka.consumer.KafkaConsumerTerminatingLifecycle is inherited by class KafkaConsumer in current version.
> [error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.awaitTermination")
> ```
> 
> because `KafkaConsumer` still exposes those methods, unless I've misunderstood what you were suggesting @bplommer. I moved the new `terminate` methods to `KafkaConsumerTerminatingLifecycle`, made it extend `KafkaConsumerLifecycle` and made `KafkaConsumer` extend `KafkaConsumerTerminatingLifecycle`

I just added the new methods to `mimaBinaryIssueFilters` as this should all be on the 1.4 branch

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
agustafson: Yeah, fair point, it would need a lot more testing. However, it does seem a little odd to not run `KafkaConsumer.stopConsuming` as part of the termination methods.

For the moment I'll just get `terminate` and `awaitTerminate` to point at the old `fiber.join/cancel` methods

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -554,6 +547,10 @@ object KafkaConsumer {
 
       override def toString: String =
         "KafkaConsumer$" + id
+
+      override def terminate: F[Unit] = fiber.cancel
```
agustafson, line 551: Perhaps this should be 
```suggestion
      override def terminate: F[Unit] = polls.cancel *> actor.cancel
```
or
```suggestion
      override def terminate: F[Unit] = polls.cancel.start.void *> actor.cancel.start.void
```

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -554,6 +547,10 @@ object KafkaConsumer {
...
+
+      override def awaitTermination: F[Unit] = fiber.join
```
agustafson, line 553: Perhaps this should be 
```suggestion
      override def terminate: F[Unit] = polls.join *> actor.join
```
or
```suggestion
      override def terminate: F[Unit] = polls.join.attempt.void *> actor.join.attempt.void
```

It just means that the shutdown is ordered. What do you think?

bplommer:
> > > > Getting mima compatibility errors because we've added new methods, but not sure how we deal with them
> 
> > > 
> 
> > > 
> 
> > > Yeah, that's a pain. My best suggestion is to add the new methods to a `trait KafkaConsumerLifecycleV2[F[_]] extends KafkaConsumerLifecycle[F]` and replace `KafkaConsumerLifecycle` with `KafkaConsumerLifecycleV2` in the definition of `KafkaConsumer`.
> 
> > 
> 
> > I still get
> 
> > 
> 
> > ```
> 
> > [error] fs2-kafka: Failed binary compatibility check against com.github.fd4s:fs2-kafka_2.13:1.3.1! Found 2 potential problems
> 
> > [error]  * abstract method terminate()java.lang.Object in interface fs2.kafka.consumer.KafkaConsumerTerminatingLifecycle is inherited by class KafkaConsumer in current version.
> 
> > [error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.terminate")
> 
> > [error]  * abstract method awaitTermination()java.lang.Object in interface fs2.kafka.consumer.KafkaConsumerTerminatingLifecycle is inherited by class KafkaConsumer in current version.
> 
> > [error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("fs2.kafka.KafkaConsumer.awaitTermination")
> 
> > ```
> 
> > 
> 
> > because `KafkaConsumer` still exposes those methods, unless I've misunderstood what you were suggesting @bplommer. I moved the new `terminate` methods to `KafkaConsumerTerminatingLifecycle`, made it extend `KafkaConsumerLifecycle` and made `KafkaConsumer` extend `KafkaConsumerTerminatingLifecycle`
> 
> 
> 
> I just added the new methods to `mimaBinaryIssueFilters` as this should all be on the 1.4 branch

That seems right. Mima warns about new methods in sealed traits even though they're not a problem.

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
bplommer: I *think* that should be ok. Only thing is that whether `cancel` blocks is implementation-dependent - it doesn't with `IO`, but it can with other implementations of the CE typeclasses.

I think the worst that could cause would be slower shutdown rather than deadlock, but again I'm not certain.

Maybe we should start with inclining the `combine` stuff from `Fiber` and go from there.

bplommer (approved):
LGTM

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
bplommer: I think that would swallow errors? Again, probably better to start by inlining the combine operation on Fiber. But I think that can be a separate PR.

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
SystemFw: > Only thing is that whether cancel blocks is implementation-dependent - it doesn't with IO, but it can with other implementations of the CE typeclasses.

It does with `ce2.IO`, but not always due to implementation quirks, it always does with `ce3.IO`.  

I don't know what the rationale is here for not awaiting for shutdown, but the reason we have await-cancel  semantics in CE is that doing `cancel.start.void` at the call site to avoid waiting is pretty easy, whereas if you don't wait, and call-site wants to, you're in trouble

bplommer: >  doing cancel.start.void at the call site to avoid waiting is pretty easy, whereas if you don't wait, and call-site wants to, you're in trouble

Can't you do `fiber.cancel *> fiber.join.void` at the call site in that case?

Either way, I agree it's probably better to have `terminate` await termination - I mainly just wanted to avoid changing both API and behaviour in the same PR. It would be interesting to hear any thoughts from @vlovgr or @LMnet on this.

SystemFw: > Can't you do fiber.cancel *> fiber.join.void at the call site in that case?


if the implementation of `cancel` does not wait for the resource to close (e.g. it's `thing.close.start.void`), and you want to wait for that at call site, there's nothing you can do as you've lost a handle on the fiber.

Also in ce2 if you do `fiber.cancel *> fiber.join` , it's going to be non-terminating, and in ce3 you're going to have an `Outcome`, but that still doesn't change anything in the general case unless `join` happens to wait for the resource closing (which I think it's the case here), but then just having one method to "cancel and wait", which you can `start`, seems easier

vlovgr: The reason why we don't use the real `Fiber#join` and `cancel` here is due to the quirky behaviours @SystemFw describe. I'm not sure why `cancel` doesn't await termination here, but I don't see a good reason for it. I'm all for changing it. 👍

bplommer: Ok, shall we change terminate to `fiber.cancel *> awaitTermination` and merge this?

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
agustafson, line 553: By "inlining the combine operation" do you mean the following @bplommer?
```suggestion
      override def terminate: F[Unit] = (actor combine polls).cancel

      override def awaitTermination: F[Unit] = (actor combine polls).join
```

bplommer: No, I meant inline the code from `cats-effect` and `cats` that implements `combine` on fibers. So I think that would be code from the `Applicative` instance for `Fiber` and the `Semigroup` instance for `Applicative`. Then `KafkaConsumer#fiber` could just be defined as `Fiber(terminate.start.void, awaitTermination)` and would be easy to completely remove for v2.

agustafson: Ok, sorry I'm not getting it @bplommer  😕 I'm assuming that `KafkaConsumer` will still be passed `polls: Fiber[F, Unit]` and `actor: Fiber[F, Unit]`? If that's the case then isn't it best to just use the `combine` functionality rather than inline it?

Will change the definition of `fiber = Fiber(terminate.start.void, awaitTermination)`

agustafson: Actually, should `val fiber = fiber = Fiber(awaitTermination.start.void, terminate)` instead, if `terminate` is cancellation and `awaitTermination` is joining?

bplommer: Yes, they'll still be passed `polls` and `actor` - these are "real" fibers produced by the CE runtime system. But `pollsFiber` and `actorFiber` are "fake" fibers, in the sense that they're created by synthesising `join` and `cancel` operations rather than by invoking `.start` on an effect. 

This (a) is kind of hacky and (b) doesn't work in CE3, where you can't synthesise fibers in the same way. (b) means that if we want the same behaviour in CE3 we'll need to implement it ourselves anyway - so for consistency across branches it's better to do it in 1.x.

bplommer: > Actually, should `val fiber = Fiber(awaitTermination.start.void, terminate)` instead, if `terminate` is cancellation and `awaitTermination` is joining?

Sorry, yes - I got the order wrong there.

agustafson: Thanks @bplommer, I've made some updates: https://github.com/fd4s/fs2-kafka/pull/504/commits/dbf05abbfd2c5fb1c8f48e629ca32b7430b2d9cc
but I didn't understand the inlining work to do. I think you're suggesting we add logic similar to https://github.com/typelevel/cats-effect/blob/series/2.x/core/shared/src/main/scala/cats/effect/Fiber.scala#L109-L117
To me I'm not sure why we don't just use the existing `combine` method.

bplommer: Yes, that's the bit of logic I'm talking about inlining.

To be honest, for the CE2-based versions of fs2-kafka I think this is an implementation detail that doesn't hugely matter. It's probably best if we keep the existing `combine` for now and then think about backporting whatever changes we make in the CE3 branch.

agustafson: Great! Thanks @bplommer.

agustafson:
I think I've addressed all the questions and issues, but let me know if there are any other concerns.

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -134,21 +127,8 @@ object KafkaConsumer {
...
-          Fiber[F, Unit](actor.join.guaranteeCase {
-            case ExitCase.Completed => polls.cancel
-            case _                  => F.unit
-          }, actor.cancel)
-
-        val pollsFiber =
-          Fiber[F, Unit](polls.join.guaranteeCase {
-            case ExitCase.Completed => actor.cancel
-            case _                  => F.unit
-          }, polls.cancel)
```
bplommer, line 148: This logic no longer exists in the updated PR but all the tests still pass - does this indicate a failure mode that we don't have a test for and should add tests? (@vlovgr)

agustafson: Good spot, there wasn't much logic in the test around the `fiber`. This is all I could find: https://github.com/fd4s/fs2-kafka/blob/series/1.x/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala#L203-L218

agustafson: As for the above block, that has been deleted, it seems a bit odd to me. I would have thought that if the `join` was successful that you shouldn't `cancel` the other `Fiber`. The `actorFiber combine pollsFiber` would `join` the `Fiber`s one after the other, so the `cancel` shouldn't be required (as far as I can tell).

bplommer:
I've updated this PR to revert the change to fiber cancelation logic, and am going to go ahead and merge it - we can think later about the other change.

PR #505, opened 2021-01-26
KafkaConsumer: use `fs2.Stream` instead of `foreverM`

agustafson:
It seemed that the consumer actor and poll scheduler were both doing stream-like things, so replaced the usage of `foreverM` with `fs2.Stream` operations.


agustafson:
`sbt ci` hangs so looks like it doesn't work

PR #509, opened 2021-02-01
KafkaConsumer signed partition stream

agadek:
my suggestion of KafkaConsumer API enhancement. Having topic partition instance tupled with a topic stream is useful for logging and metrics. I know it's doable via 'partitionsMapStream' but it'd be nice to have it out of the box

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -333,6 +333,15 @@ object KafkaConsumer {
         }
       }
 
+      override def signedPartitionedStream: Stream[F, (TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]])] = {
+        partitionsMapStream.flatMap { partitionsMap =>
+          Stream.emits(partitionsMap.toVector.map {
+            case (topicPartition, partitionStream) =>
+              topicPartition -> partitionStream
+          })
+        }
+      }
+
       override def stream: Stream[F, CommittableConsumerRecord[F, K, V]] =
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -333,6 +333,15 @@ object KafkaConsumer {
         }
       }
 
+      override def signedPartitionedStream: Stream[F, (TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]])] = {
+        partitionsMapStream.flatMap { partitionsMap =>
+          Stream.emits(partitionsMap.toVector.map {
+            case (topicPartition, partitionStream) =>
+              topicPartition -> partitionStream
+          })
```
bplommer, line 341: Can't this just be the following?
```suggestion
          Stream.emits(partitionsMap.toVector)
```

agadek: I know I can but IMO still missing it in API

bplommer:
Thanks for this contribution! I like this but I'm not sure about the name - maybe `labelledPartitionedStream`?

Also, we can't add new abstract methods to `KafkaConsume` for binary compatibility reasons, but we can add them to `KafkaConsumer` because it's sealed. I suggest defining the new method directly in `KafkaConsumer` and then we can add a new trait later.

agadek:
will do!

aartigao:
Closing this PR because targets `series/1.x` and we're on `series/3.x` (also last activity was +2 years ago).

Thanks for your time and feel free to open it again if needed 🙏🏽

PR #511, opened 2021-02-03
correct deprecation message for adminClientStream

jopecko:
noticed that the deprecation warning didn't match what change was required to fix it


bplommer:
Thanks!

PR #514, opened 2021-02-08
Fix consumer lifecycle bincompat

bplommer:
I made a bincompat mistake with #504 - this fixes it by providing a default implementation of the new methods in `KafkaConsumerLifecycle`


bplommer:
This is running into what looks like a bug with `@nowarn` in scala 2.12 - will dig in to it.

PR #573, opened 2021-04-01
More idiomatic syntax for instantiating things via partial application

### LINKED ISSUE #571: Simplify syntax for instantiating consumers etc
Currently we have `KafkaConsumer.resource[F].using(settings)` etc. Instead this can be `KafkaConsumer[F].resource(settings)`. This is simpler and more idiomatic, and let’s us get rid of classes like `KafkaConsumerResource` which, confusingly, isn’t a resource.

I’ll have a PR for this shortly.

bplommer:
ScalaFix to follow. Resolves #571


LMnet (approved):
:+1:

Issue #553, opened 2021-03-21
Separate initialization effect in the constructors


LMnet:
Many of the components of the library have the constructor in the following form:
```scala
class Component[F]
object Component {
  def apply[F[_]: SomeConstraints]: F[Component[F]]
}
```

The problem with this form of a constructor is that the initialization effect and execution effect are the same effects. But in fact, they may be different. Moreover, they may have different constraints.

I suggest to change in 2.0.0 all these constructors to the following form:
```scala
class Component[F]
object Component {
  def apply[F[_]: ExecConstraints, G[_]: InitConsraints]: G[Component[F]]
}
```

This change will be especially helpful for the complex tagless final codebases which use multiple different reader-like effects in the same codebase.

bplommer:
Just to be clear - you're talking about allocating `KafkaConsumer`/`KafkaProducer` etc as `Resource[G, KafkaConsumer[F, K, V]]` etc, right?

LMnet:
Yes, exactly.

catostrophe:
I would also keep a form of this method with the same effects passed, e.g.
```scala
def make[F[_]: ExecConstraints, G[_]: InitConsraints]: G[Component[F]] = ???
def apply[F[_]: ExecConstraints: InitConsraints]: F[Component[F]] = make[F, F] 
```

bplommer:
Agreed. I think we should keep the current methods as they are and add variants with names suffixed with `In` with the extra type parameter.

bplommer:
I’ve put this in the v3.0.0 milestone, but hopefully we can do it in a non-breaking way as part of the 2.x series.

bplommer:
This is done for `KafkaAdminClient` and for `KafkaProducerConnection`, with the caveat that for `KafkaProducerConnection` it requires `Async` for the allocating effect purely so that `close` can be run on a custom ExecutionContext if provided.

Doing this for `KafkaConsumer` is going to be much harder, if it's possible at all, because the actor is initialized and terminated as part of the allocation.

TODO:
[ ] `KafkaProducer`
[ ] `TransactionalKafkaProducer`
[ ] Overloads that only require `Sync` for allocation, and always run `close` on the default blocking pool

PR #581, opened 2021-04-05
Decouple resource instantiation from settings

bplommer:
This is binary-breaking but mostly source-compatible, so it seems worth doing before 2.0.0 given it's likely to be needed for #553.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/AdminClientSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/AdminClientSettings.scala
@@ -20,7 +20,7 @@ import scala.concurrent.duration._
   * [[AdminClientSettings]] instances are immutable and all modification
   * functions return a new [[AdminClientSettings]] instance.<br>
   * <br>
-  * Use [[AdminClientSettings.apply]] for the default settings, and
+  * Use [[AdminClientSettings#apply]] for the default settings, and
   * then apply any desired modifications on top of that instance.
   */
 sealed abstract class AdminClientSettings {
--- a/modules/core/src/main/scala/fs2/kafka/AdminClientSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/AdminClientSettings.scala
@@ -6,10 +6,8 @@
 
 package fs2.kafka
 
-import cats.effect.Sync
 import cats.Show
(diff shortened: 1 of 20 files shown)


modules/core/src/main/scala/fs2/kafka/admin/MkAdminClient.scala
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package fs2.kafka.admin
+
+import cats.effect.Sync
+import fs2.kafka.AdminClientSettings
+import org.apache.kafka.clients.admin.AdminClient
+import fs2.kafka.internal.converters.collection._
+
+trait MkAdminClient[F[_]] {
```
LMnet, line 14: Maybe at least a minimal description about it in scaladoc should be added?

modules/core/src/main/scala/fs2/kafka/AdminClientSettings.scala
```diff
@@ -25,7 +23,7 @@ import scala.concurrent.duration._
   * Use [[AdminClientSettings#apply]] for the default settings, and
   * then apply any desired modifications on top of that instance.
   */
-sealed abstract class AdminClientSettings[F[_]] {
+sealed abstract class AdminClientSettings {
```
LMnet, line 26: With this change, I wonder why we can't use common case class here? The current solution provides almost the same functionality but with tons of boilerplate inside the library.

LMnet (changes requested):

modules/core/src/main/scala/fs2/kafka/AdminClientSettings.scala
bplommer: I agree it's boilerplatey,but the problem with using case classes directly for config is that adding fields breaks binary compatibility.

One thing we could do to reduce boilerplate, though, is implement the `withProperty`-based methods directly in the abstract class rather than in the implementation class.

LMnet: Oh, I forgot about this problem with case classes.

vlovgr (approved):

LMnet (approved):

Issue #582, opened 2021-04-08
Use single-threaded execution context for consumers


bplommer:
The Java `Consumer` isn't safe for concurrent access by multiple threads. `KafkaConsumerActor` *should* prevent this from happening, but there have been reports of cases where it doesn't when using a non-default blocker with fs2-kafka 1.x. To guard against this we should use a single-threaded ExecutionContext for blocking operations on he consumer in v2.x. We should also consider deprecating `withBlocker` in v1.x, since a single-threaded blocker should always be used.

bplommer:
Resolved by #584

PR #583, opened 2021-04-08
Update doc of ConsumerSettings#withBlocker

bplommer:
See #582

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -51,19 +51,20 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
 
   /**
     * The `Blocker` to use for blocking Kafka operations. If not
-    * explicitly provided, a default `Blocker` will be created
-    * when creating a `KafkaConsumer` instance.
+    * explicitly provided, a default single-threaded `Blocker`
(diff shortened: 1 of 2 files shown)


LMnet:
Do we really need to deprecate it? Maybe we could just add some additional information about requirements about thread-pool in the scaladoc. Because custom thread-pool may be useful for custom error handling, for logging, or to have good thread names.

bplommer:
> Do we really need to deprecate it? Maybe we could just add some additional information about requirements about thread-pool in the scaladoc. Because custom thread-pool may be useful for custom error handling, for logging, or to have good thread names.

I don't have a particularly strong opinion on this. It does raise another question though - if this is an ability we want to keep, should we also allow users of 2.x to provide custom ExecutionContexts for blocking operations?

LMnet:
> should we also allow users of 2.x to provide custom ExecutionContexts for blocking operations?

I think we should.

bplommer:
> I think we should.

I'm fine with that. So we'd have `withCustomBlockingContext(ec: ExecutionContext)` in the `Settings` classes (I think the word `custom` should be there because it's confusing if `blockingContext` can be `None` when a blocking context is always required). The fact that we have a `Blocking` abstraction should make it pretty easy to do, as in #584.

LMnet:
Sounds good for me.

bplommer:
I've revised this PR to just update the doc - I'll go ahead and merge.

PR #585, opened 2021-04-08
Add missing MkProducer argument, force compiler errors when they're missing

bplommer:
I'm not 100% happy with the added boilerplate, or with it being needed, but I don't have a better idea. Any thoughts?

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaAdminClient.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaAdminClient.scala
@@ -577,10 +577,10 @@ object KafkaAdminClient {
     * working in a `Stream` context, you might instead prefer to
     * use the [[KafkaAdminClient.resource]].
     */
-  def stream[F[_]: Async: MkAdminClient](
+  def stream[F[_]](
     settings: AdminClientSettings
-  ): Stream[F, KafkaAdminClient[F]] =
-    Stream.resource(KafkaAdminClient.resource(settings))
+  )(implicit F: Async[F], mk: MkAdminClient[F]): Stream[F, KafkaAdminClient[F]] =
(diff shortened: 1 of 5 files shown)


bplommer:
Exciting update: it doesn't compile with Scala 2.12 anyway. I guess I'll just leave the ambiguous implicits stuff out, but I'm still not happy with that.

bplommer:
Exciting update no. 2: It now compiles in Scala 2.12, courtesy of added ugliness.

LMnet:
Could you please explain why do we need this? As far as I understand, without `mkAmbig1` and `mkAmbig2` default instance of the `Mk???` instance will be used (which is derived from the `Sync`). And what is the problem with that?

bplommer:
> Could you please explain why do we need this? 

It's not strictly needed, but it's an extra safeguard against us writing factory methods that don't require an `Mk` argument.

> As far as I understand, without `mkAmbig1` and `mkAmbig2` default instance of the `Mk???` instance will be used (which is derived from the `Sync`). And what is the problem with that?

There's no problem with the default instance derived from `Sync` being used, but that derivation needs to happen at the call site in user code so that the user can override the default behaviour if needed.

LMnet:
Oh, so that's for library maintainers mostly, not for users. Ok, I got it.

LMnet (approved):

bplommer:
This is intended to catch bugs like the one this PR fixes on lines 642-63 of `KafkaProducerConnection.scala`, where the `stream` method delegates to the `resource` method. Rather than `stream` taking a `MkProducer` argument it was autoderiving an instance based on `Sync`, which would prevent use of a custom instance by user code.

bplommer:
> Oh, so that's for library maintainers mostly, not for users. Ok, I got it.

👍 didn't see your reply before my second comment

PR #608, opened 2021-05-02
Demonstrate issues with Vulcan deserializer

bplommer:
This is to illustrate some issues with the vulcan deserializer not behaving as intended - I'm going to open an issue shortly.


aartigao:
Ben, after so many years I think it's time to do some cleanup and close old drafts to avoid noise in the Pull Request section and focus on newer things.

OFC, feel free to reopen it if you think it's worth 👍🏽

PR #629, opened 2021-06-09
Vulcan testkit

keirlawson:
Per discussion on gitter, an munit fixture to facilitate testing avro schema compatibility against the schema registry.  Feedback welcome :)


keirlawson:
Is it possible to re-run CI? I initially targetted the wrong branch which is the cause of the test failures

keirlawson:
Not sure why this build is failing?

keirlawson:
SO far as I can make out the issue is that this is a new module so mima is not able to check binary compat against a previous version, not sure how best to get around this

bplommer:
The new module needs to be temporarily excluded from Mima. There’s probably a cleaner way to do it but in the meantime I’m happy to do it myself the hacky way.

modules/vulcan-testkit/src/main/scala/fs2/kafka/vulcan/testkit/SchemaSuite.scala
```diff
@@ -0,0 +1,76 @@
...
+
+trait SchemaSuite extends FunSuite {
+
+  private def codecAsSchema[A](codec: Codec[A]) = codec.schema.fold(e => fail(e.message), ok => ok)
+
+  def compatibilityChecker(
+    clientSettings: SchemaRegistryClientSettings[IO],
+    name: String = "schema-compatibility-checker"
+  ) = new Fixture[CompatibilityChecker[IO]](name) {
+    private var checker: CompatibilityChecker[IO] = null
```
bplommer, line 31: Couldn't this be a lazy val rather than a var initialized in `beforeAll`?

keirlawson: I guess that might mess with test timings then, as the first test to use the fixture would have the spin-up time bundled in?

keirlawson: looking into the effect in question, don't think its actually doing any IO, so shouldn't add too much overhead to the first test to use it, that said I think I still find spinning up in the beforeAll slightly neater, even if the presence of `null` in the codebase makes me feel slightly ill...

bplommer: Yeah, it probably doesn't matter much either way.

bplommer:
Sorry to be so slow with this! I'm going to take the liberty of splitting this into `vulcan-testkit` and `vulcan-testkit-munit` modules, then we should be good to go.

bplommer:
On second thought, let's just leave this as one module called vulcan-testkit-munit unless and until we want to add another test framework integration.

Issue #446, opened 2020-12-02
Convenience methods on `KafkaProducer`


bplommer:
- `produce_(producerRecords: ProducerRecords[K, V, _]): F[F[Chunk[(ProducerRecord[K, V], RecordMetadata)]]]`
- `produceOne[P](producerRecord: ProducerRecord[K, V], passthrough: P)`: F[F[ProducerResult[K, V, P]]]
- `produceOne[P](topic: String, key: K, value: V, passthrough: P): F[F[ProducerResult[K, V, P]]]`
- `produceOne_(producerRecord: ProducerRecord[K, V]): F[F[RecordMetadata]]`
- `produceOne_(topic: String, key: K, value: V): F[F[RecordMetadata]]`

PR #632, opened 2021-06-14
Add producer convenience methods

keirlawson:
Per suggestions in #446 

Tests are def a bit anaemic, wasn't sure what the right balance between exhaustivity and maintainability was.

Didn't include a `pipe` method as suggested in the original issue as was not clear what the semantics should be (flatten the `F[F[...]]`?), but happy to add if clarified.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
@@ -58,7 +58,8 @@ abstract class KafkaProducer[F[_], K, V] {
 
 object KafkaProducer {
 
-  implicit class ProducerOps[F[_], K, V](private val producer: KafkaProducer[F, K, V]) extends AnyVal {
+  implicit class ProducerOps[F[_], K, V](private val producer: KafkaProducer[F, K, V])
+      extends AnyVal {
 
     /**
       * Produce a single [[ProducerRecord]] without a passthrough value,
--- a/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
@@ -58,13 +58,13 @@ abstract class KafkaProducer[F[_], K, V] {
 
 object KafkaProducer {
 
-  implicit class ProducerOps[F[_]: Functor, K, V](producer: KafkaProducer[F, K, V]) {
+  implicit class ProducerOps[F[_], K, V](private val producer: KafkaProducer[F, K, V]) extends AnyVal {
 
     /**
       * Produce a single [[ProducerRecord]] without a passthrough value,
       * see [[KafkaProducer.produce]] for general semantics.
       */
-    def produceOne_(record: ProducerRecord[K, V]): F[F[RecordMetadata]] =
+    def produceOne_(record: ProducerRecord[K, V])(implicit F: Functor[F]): F[F[RecordMetadata]] =
       produceOne(record, ()).map(_.map { res =>
         res.records.head.get._2 //Should always be present so get is ok
       })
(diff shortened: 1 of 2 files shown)


bplommer:
This looks great! Just one thing - adding these as instance methods is going to break binary compatibility, so I'm thinking it would be better to have them as extension methods implemented in terms of `produce`. Would that work?

keirlawson:
Should be do-able, where would you place said extension methods?

bplommer:
> Should be do-able, where would you place said extension methods?

I'd say an implicit class in companion object so they're always available. It'll need some kind of typeclass available at the usage site (I think `Applicative` or `Monad`?) but that's ok.

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -56,6 +58,53 @@ abstract class KafkaProducer[F[_], K, V] {
...
+  implicit class ProducerOps[F[_] : MonadThrow, K, V](producer: KafkaProducer[F, K, V]) {
+
+      /**
+        * Produce a single [[ProducerRecord]] without a passthrough value,
+        * see [[KafkaProducer.produce]] for general semantics.
+        */
+      def produceOne_(record: ProducerRecord[K,V]): F[F[RecordMetadata]] = {
+        produceOne(record, ()).map(_.flatMap { res =>
+          val metadata = res.records.head.map(_._2)
+          MonadError[F,Throwable].fromOption(metadata, new IllegalStateException("Exactly one record metadata expected"))
```
bplommer, line 70: Is this the only reason we need `MonadThrow`? If so I'd be inclined to just throw an exception here, since this should never happen.

bplommer:
> Didn't include a `pipe` method as suggested in the original issue as was not clear what the semantics should be (flatten the `F[F[...]]`?), but happy to add if clarified.

Sorry, forgot about this - I was thinking it would be the same as the existing `pipe` but as an extension method: https://github.com/fd4s/fs2-kafka/blob/a3832589e6e85c2d21b67a0e146809b35f06ceb9/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L171

That requires `ProducerSettings` though, which would be weird on an extension method, so let's leave it for now 🤔

bplommer:
One other thing I forgot - can you make the ops class extend `AnyVal`?

keirlawson:
> One other thing I forgot - can you make the ops class extend `AnyVal`?

Doing so I get `value class needs to have exactly one val parameter`, presumably because the context bound desugars to an implicit parameter?

bplommer:
iirc the `Producer` parameter should be a `private val`.

keirlawson:
> iirc the `producer` parameter should be a `private val`.

I get the same error with that

LMnet:
> Doing so I get value class needs to have exactly one val parameter, presumably because the context bound desugars to an implicit parameter?

Yes, to use `extends AnyVal` you have to move `Functor` requirement to the methods, as an implicit parameter.

Also, I would add tests for each method. These tests could be dumb, but each new method should be called in tests. Because these multiple overrides could lead to runtime issues with erasure, especially after refactorings.

keirlawson:
> > > Doing so I get value class needs to have exactly one val parameter, presumably because the context bound desugars to an implicit parameter?
> > 
> > 
> > Yes, to use `extends AnyVal` you have to move `Functor` requirement to the methods, as an implicit parameter.
> 
> Right, yeah. On second thought, maybe these should just be `final` instance methods implemented in the trait - what do you guys think?

Personally not strong opinions, happy to do whichever folk reckon is best :)

bplommer:
> Also, I would add tests for each method. These tests could be dumb, but each new method should be called in tests. Because these multiple overrides could lead to runtime issues with erasure, especially after refactorings.

I think erasure issues should show up at compile time? But yes, it would be good to at least see that code using these compiles.

LMnet:
> On second thought, maybe these should just be final instance methods implemented in the trait - what do you guys think?

I'm on the `Ops` side. I think basic interface should be as simple as possible. And all custom stuff on top of it could be done in the separate `Ops` trait.

bplommer:
> > On second thought, maybe these should just be final instance methods implemented in the trait - what do you guys think?
> 
> 
> 
> I'm on the `Ops` side. I think basic interface should be as simple as possible. And all custom stuff on top of it could be done in the separate `Ops` trait.

I don't have a strong opinion so let's go with that 🙂

bplommer (approved):
LGTM - thanks!

LMnet (approved):

PR #643, opened 2021-07-16
Add a minimal readme

### LINKED ISSUE #642: Add a readme
- [ ] Getting started
- [ ] Choosing a version
- [ ] Adopters
- [ ] Related libraries 
- [ ] Contributing

bplommer:
It could use some fleshing out, but it's a start.

Resolves #642

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,31 @@
+# FS2-Kafka: Functional Kafka Streams for Scala
+
+[![Latest version](https://index.scala-lang.org/fd4s/fs2-kafka/fs2-kafka/latest.svg?style=flat)](https://index.scala-lang.org/fd4s/fs2-kafka/fs2-kafka)
+![License](https://img.shields.io/github/license/fd4s/fs2-kafka)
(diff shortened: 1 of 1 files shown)


vlovgr:
How about instead adding the new sections and other details to the [overview](https://fd4s.github.io/fs2-kafka/docs/overview)? My preference would be to have everything on the microsite and just link there from the readme. What do you think?

(Just a note about the downside with having sbt dependency coordinates in the readme like this: we manually have to keep them updated, whereas on the microsite it's generated based on the lastest released version. It can be solved by also generating the readme, but not sure I'd like to go down that road.)

SystemFw:
These days I prefer very minimal readmes as well https://github.com/systemfw/dynosaur

bplommer:
Fair points, but I do still think it's worth including a ~1-page overview to function as a tl;dr for the microsite with links to specific sections.  It feels more welcoming to me, and it's one less context switch that people probably don't need by the time they've resorted to looking at docs 😉

LMnet (approved):

PR #661, opened 2021-08-18
Add build.sbt instructions re: fs2-kafka-vulcan

### LINKED ISSUE #660: Adding dependency on fs2-kafka-vulcan fails to resolve
When you add the dependency on fs2-kafka-vulcan to your project, it won't resolve unless you have the confluent resolver added;
```scala
resolvers += "confluent" at "https://packages.confluent.io/maven/",
```

Otherwise, you get

> ResolveException: Error downloading io.confluent:kafka-avro-serializer:6.1.1

This should be added to the docs for the vulcan integration

Daenyth:
Close #660


vlovgr:
Thanks! Although this is actually already documented [here](https://fd4s.github.io/fs2-kafka/docs/overview#modules).

bplommer:
I think it's still worth having - it's already tripped up at least one person...

PR #662, opened 2021-08-19
Extension methods for operating directly on `Stream` of `KafkaConsumer`

bplommer:


docs/src/main/mdoc/consumers.md
```diff
@@ -200,8 +200,8 @@ object ConsumerStreamExample extends IOApp {
   def run(args: List[String]): IO[ExitCode] = {
     val stream =
       KafkaConsumer.stream(consumerSettings)
-        .evalTap(_.subscribeTo("topic"))
-        .flatMap(_.stream)
+        .subscribeTo("topic")
+        .stream
```
bplommer, line 204: I feel like maybe we should rename `stream` to `records` to avoid having common patterns with `stream` being used twice with different meanings.

LMnet: Sounds reasonable to me. We could try to add new methods in a 3.0 without deletions of old methods.

LMnet (approved):
:+1:

PR #664, opened 2021-08-20
Alias `stream` as `records`

bplommer:
I think this improves clarity, especially with the extension methods on allocating streams. I'd like to get this in for the next release, as having two uses of `.stream` with totally different meanings looks weird to me.

I don't see why we can't do this in the 1.x/2.x series - we can consider later whether we want to deprecate `stream` at some time in the future.

### DIFF
--- a/docs/src/main/mdoc/consumers.md
+++ b/docs/src/main/mdoc/consumers.md
@@ -201,7 +201,7 @@ object ConsumerStreamExample extends IOApp {
     val stream =
       KafkaConsumer.stream(consumerSettings)
         .subscribeTo("topic")
-        .stream
+        .records
 
     stream.compile.drain.as(ExitCode.Success)
   }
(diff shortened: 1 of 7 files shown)


LMnet:
Sorry for the late review, was a bit busy.

I want to suggest another approach for a new method. At the current moment we have `stream`, `partitionedStream` and `partitionsMapStream`. All these methods could be used for retrieving records from a Kafka. I suggest grouping these methods into something like this:
```scala
consumer.records.flat          // consumer.stream
consumer.records.partitioned   // consumer.partitionedStream
consumer.records.partitionsMap // consumer.partitionsMapStream (here we can try to find a better name)
```

fs2 uses the same approach for a `compile` methods.

I think this approach will make API more discoverable, and it will give us more space for adding new options of records retrieving.

Issue #663, opened 2021-08-20
Improve resilience of serialisation using Vulcan amd Schema Registry


gordon-rennie:
Serialisation / deserialisation using vulcan codecs in `fs2-kafka-vulcan` requires communication with a Schema Registry REST API, implemented in Confluent's Java code which `fs2-kafka-vulcan` wraps. Transient problems communicating with the registry's API - e.g. network problems, timeouts, unexpected server-side errors - bubble up from the schema registry client's Java code to `fs2-kafka-vulcan` and cause fatal exceptions unless deliberately handled in your application code - `fs2-kafka-vulcan` makes no attempts to retry. This makes Avro ser/des a weak point for `fs2-kafka` application resiliency.

For example, I recently experienced an unhealthy Kafka cluster affecting both broker and schema registry. Interactions with the broker were robust and self-recovered from various errors, but a `SocketTimeoutException` communicating with the Schema Registry during serialisation blew up the application ([stacktrace](https://gist.github.com/gordon-rennie/923091c0504eaca993be5809629b24fe)).

Is there appetite to include Schema Registry retries in `fs2-kafka-vulcan`? As I see it, ideally the library would have some default retry config - similar to the situation with the Kafka broker publisher/consumer. Having reviewed the Confluent Java code, there's no allowance for retries in the client or related classes, so this would have to be added in `fs2-kafka-vulcan`.

gordon-rennie:
I have done some code analysis for serialisation:
* `fs2-kafka-vulcan` interfaces with the Java code here:
https://github.com/fd4s/fs2-kafka/blob/de26d891bdf4f28585371d7e480ef99c324010f2/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSerializer.scala#L26
* `F.pure(serializer.serialize(topic, value))` doesn't seem quite right to me - `serializer.serialize` calls into the Java code, which is not pure and performs IO including checking a schema cache and (on cache miss) making a REST API call. This seems like the call site that should be enhanced with retry logic.
[...]

Issue #623, opened 2021-05-27
Rework or remove CommitRecovery


LMnet:
Recently I have worked with the `CommitRecovery` functionality. And I found that this part of the library is not pleasant to work with. 

In particular, I faced the next issues:
1. `recoverCommitWith` signature has specific bounds on `F`. And there is no way to add some extra bounds if I want. For example, I wanted to use log4cats and add some details about retrying inside `recoverCommitWith`. And at the current moment, I can't that because `recoverCommitWith` doesn't have a `Sync` bound. Overall  `recoverCommitWith`'s signature feels weird and limiting.
2. `CommitRecovery.Default` is pretty nice, but not configurable.
3. I'm using pureconfig for my configuration and I had to create a custom wrapper around `CommitRecovery` to make it works nicely with the pureconfig.

Also, It's debatable that this functionality should be built-in inside fs2-kafka. For me, it looks like all this retries stuff could be delegated to a specialized solution like cats-retry. And it may look more natural and composable. ATM this part is not composable: I had to overcome some difficulties to use `CommitRecovery` with the pureconfig, log4cats, and cats-retry.

I can suggest 2 ways of solving this:
1. Completely remove this functionality. If the user will want to retry commit failures he can use any library and add retries on top of the `commit` method. As a bonus point here we will be able to remove Jitter from the library, because it uses only in `CommitRecovery` logic.
2. Somehow rework this functionality. Maybe remove `CommitRecovery` logic from the `KafkaConsumerActor` and offer a `CommitRecovery` as an additional option on top of a `commit` method?

bplommer:
Sorry, I only just saw this - had my github notifications set up wrong  🤦‍♂️ I'll try give this some thought in the next few days.

bplommer:
I feel like removing commit recovery as a default behaviour would be a regression in terms of robustness and approachability, so I'd like to avoid that if possible.

What if we convert `CommitRecovery` to be tagless final, so the bounds on `F` can be implementation-specific? It would mean you need `Sync` to instantiate `ConsumerSettings` with the default `CommitRecovery` but I don't see that as a huge problem. I think that should address your composability issue, though I'm not sure if it would address the PureConfig issue.

> 2. Somehow rework this functionality. Maybe remove `CommitRecovery` logic from the `KafkaConsumerActor` and offer a `CommitRecovery` as an additional option on top of a `commit` method?

[...]

PR #673, opened 2021-08-26
Add retries for schema registry errors

gordon-rennie:
Initial PR for feedback - aims to fix https://github.com/fd4s/fs2-kafka/issues/663

## Goal
Prevent transient schema registry API errors from blowing up Vulcan ser/des by adding retries with sensible default config.

## Implementation
* some code analysis was require to determine when to retry:
  * schema registry client operations can throw `IOException` or `RestClientException`
  * for serialization, the relevant Java [KafkaAvroSerializer](https://github.com/confluentinc/schema-registry/blob/4cbf07a513f3a2fe999088abb5a74d544707c906/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerializer.java#L141-L147) error handling maps both `IOException` and `RestClientException` (if non-client error code) to `SerializationException`
  * for deserialization, the [vulcan code](https://github.com/fd4s/fs2-kafka/blob/0cde0b0e8d227fa115271598dc4fb45102fa1c7c/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroDeserializer.scala#L32-L33) makes a call using the schema registry client itself and further calls are made by the underlying Java code ([1](https://github.com/fd4s/fs2-kafka/blob/0cde0b0e8d227fa115271598dc4fb45102fa1c7c/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroDeserializer.scala#L40), [2](https://github.com/confluentinc/schema-registry/blob/4cbf07a513f3a2fe999088abb5a74d544707c906/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroDeserializer.java#L106-L107), [3](https://github.com/confluentinc/schema-registry/blob/4cbf07a513f3a2fe999088abb5a74d544707c906/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroDeserializer.java#L305-L320)) which are mapped again to `SerializationException`
  * tl;dr: retry on `IOException`, `RestClientException`, `SerializationException` 
* added retry logic for the above exceptions
* added config for number of retries and backoff duration
  * default config is 5 retries with exponential backoff starting at 50 millis.

## Notes
* Adding backoff to retries requires `Temporal[F]`, which requires upgrading from `F: Sync` to `F: Async`. Presumably this is considered a breaking change and so this feature needs to start targetting 3.x instead? I'm new to library code, just let me know :smile: 
* There is no provision for logging in `fs2-kafka-vulcan` (there's only a private, sealed interface in `fs2-kafka` core) so I have not added any
* Retry logic is hand-crafted rather than using a dependency like `cats-retry`
* I was influenced by discussions in issue https://github.com/fd4s/fs2-kafka/issues/623 not to create any new retry interfaces and not to use `Jitter[F]`, which we seem to want to move away from
  * from the discussions in this issue, I think we would prefer a more generic interface like `def schemaRegistryRetryPolicy[A]: (F[A], Throwable) => F[A]` - this would give good flexibility for clients and allow interop with libraries like `cats-retry`. I really like this idea. Unfortunately I can't see a way to make this work: deserialization will return an arbitrary `F[A]` and `A` is basically a wildcard type that needs to accommodate any arbitrary `vulcan.Codec[A]`. Wildcard types aren't supported for anonymous functions, e.g. I can't do: 
```scala
  private[this] final case class AvroSettingsImpl[F[_]](
    ...
    schemaRegistryRetryPolicy: (F[*], SerializationException) => F[*]
```
  feedback on the interface very welcome!

## TODO
* agree/finalize interface changes
* unit test


gordon-rennie:
(rebasing from 2.x to 3.x had issues and conflicts, I just applied the diff patch as a patch instead before force-pushing but GitHub hasn't liked that. I will re-raise the PR against 3.x)

PR #683, opened 2021-09-05
Add retries for schema registry errors

### LINKED ISSUE #663: Improve resilience of serialisation using Vulcan amd Schema Registry

gordon-rennie:
Fixes #663. Updated from abandoned PR #673: rebased to target 3.x and improved retry interface. Restructured PR description for readability

## Goal
Prevent transient schema registry API errors from blowing up Vulcan serialisation/deserialisation by adding retries with sensible default config.

## Implementation
* add a new `trait SchemaRegistryClientRetry[F[_]]` allowing a retry strategy on an arbitrary computation that uses the schema registry, `F[A]`, to be specified.
* add default implementation that retries on schema registry exceptions identified in the underlying Java code (see "What `Throwable`s should we retry?" below).
  * the default strategy uses a retry time backoff, requiring `Sync` to be upgraded to `Async` on some `AvroSettings` constructors; therefore this change is breaking and targets `3.x`

## Retry Interface Design
Feedback is very welcome here. I took care in designing this, drawing significant influence from discussions on issue https://github.com/fd4s/fs2-kafka/issues/623 about commit retry.

### Proposed Interface
```scala
trait SchemaRegistryClientRetry[F[_]] {

  def withRetry[A](action: F[A]): F[A]
}

// AvroSettings.scala
sealed abstract class AvroSettings[F[_]] {
  ...
  def schemaRegistryClientRetry: SchemaRegistryClientRetry[F]
}
```
**Benefits:** very flexible and composable; users can bring their own `F` to perform logging etc. Easy interop with e.g. `cats-retry` library. Robust against future implementation changes in Vulcan or the underlying Java libraries; the retry can be applied to any computation thanks to the type param.
**Cons:** required an intermediate trait to get the type parameter `[A]` because an anonymous function can't have a type param, e.g. we can't do `def schemaRegistryClientRetry: F[A] => F[A]`.

To try and get rid of the need of the type param and new trait, I also considered a variation of this proposal where the three Vulcan call sites that use the schema registry (directly or indirectly) could take one non-type-paramaterised anonymous function each as retry strategies (e.g. `def schemaRegistryGetSchemaRetry: F[ParsedSchema] => F[ParsedSchema]`), but this would leave the retry config extremely coupled to the implementation details of the existing ser/des. 

### Alternative 1
(this was my first iteration)
```scala
// AvroSettings.scala
sealed abstract class AvroSettings[F[_]] {
  ...
  def schemaRegistryRetryMaxAttempts: Int
  def schemaRegistryRetryBackoffStrategy: Int => FiniteDuration
}
```
**Benefits:** simple interface with no new traits.
**Cons:** inflexible; no provision to add logging etc. No `cats-retry` interop.

### Alternative 2
Implement tagless final traits as facades over the relevant Java schema registry classes, allow retries to be specified on their implementations.

**Cons:** this requires huge change.

## What `Throwable`s should we retry?
Some code analysis was required to determine where the schema registry is used and what exceptions it can throw. I wanted to avoid retrying on every error, which would include deterministic failures.
  * in general, schema registry client operations can throw `IOException` or `RestClientException`
  * for serialization, the relevant Java [KafkaAvroSerializer](https://github.com/confluentinc/schema-registry/blob/4cbf07a513f3a2fe999088abb5a74d544707c906/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerializer.java#L141-L147) error handling maps both `IOException` and `RestClientException` (if non-client http error code) to `SerializationException`
  * for deserialization, the [vulcan code](https://github.com/fd4s/fs2-kafka/blob/0cde0b0e8d227fa115271598dc4fb45102fa1c7c/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroDeserializer.scala#L32-L33) makes a call using the schema registry client itself and further calls are made by the underlying Java code ([1](https://github.com/fd4s/fs2-kafka/blob/0cde0b0e8d227fa115271598dc4fb45102fa1c7c/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroDeserializer.scala#L40), [2](https://github.com/confluentinc/schema-registry/blob/4cbf07a513f3a2fe999088abb5a74d544707c906/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroDeserializer.java#L106-L107), [3](https://github.com/confluentinc/schema-registry/blob/4cbf07a513f3a2fe999088abb5a74d544707c906/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroDeserializer.java#L305-L320)) which are mapped again to `SerializationException`

TL;DR this new helper function catches the relevant exceptions:
```scala
  val isRetriable: Throwable => Boolean = {
    case _: SerializationException     => true
    case _: IOException                => true
    case apiError: RestClientException => apiError.getErrorCode >= 500
    case _                             => false
  }
```
I made the function public so users of `cats-retry` etc. have it available to them for easy re-use.

## TODO
Once the interface is agreed, I need to add unit tests and docs.

### DIFF
--- a/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSettings.scala
+++ b/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSettings.scala
@@ -6,8 +6,9 @@
 
 package fs2.kafka.vulcan
 
-import cats.effect.Sync
+import cats.effect.Async
 import cats.implicits._
+
 import scala.jdk.CollectionConverters._
 import fs2.kafka.internal.syntax._
 
@@ -80,6 +81,22 @@ sealed abstract class AvroSettings[F[_]] {
   def withProperties(properties: Map[String, String]): AvroSettings[F]
 
   /**
+    * TODO
+    */
+  def schemaRegistryClientRetry: SchemaRegistryClientRetry[F]
+
+  /**
+    * TODO
+    *
+    * Creates a new `AvroSettings` instance with the specified
+    * retry strategy when recoverable Schema Registry client
+    * errors are experienced during serialization/deserialization.
+    */
+  def withSchemaRegistryClientRetry(
+    schemaRegistryClientRetry: SchemaRegistryClientRetry[F]
+  ): AvroSettings[F]
+
+  /**
     * Creates a new `KafkaAvroDeserializer` using the settings
     * contained within this [[AvroSettings]] instance, and the
     * specified `isKey` flag, denoting whether a record key or
@@ -126,8 +143,9 @@ object AvroSettings {
     override val properties: Map[String, String],
     // format: off
     val createAvroDeserializerWith: (F[SchemaRegistryClient], Boolean, Map[String, String]) => F[(KafkaAvroDeserializer, SchemaRegistryClient)],
-    val createAvroSerializerWith: (F[SchemaRegistryClient], Boolean, Map[String, String]) => F[(KafkaAvroSerializer, SchemaRegistryClient)]
+    val createAvroSerializerWith: (F[SchemaRegistryClient], Boolean, Map[String, String]) => F[(KafkaAvroSerializer, SchemaRegistryClient)],
     // format: on
+    override val schemaRegistryClientRetry: SchemaRegistryClientRetry[F]
   ) extends AvroSettings[F] {
     override def withAutoRegisterSchemas(autoRegisterSchemas: Boolean): AvroSettings[F] =
       withProperty("auto.register.schemas", autoRegisterSchemas.toString)
@@ -150,6 +168,10 @@ object AvroSettings {
 
     override def withProperties(properties: Map[String, String]): AvroSettings[F] =
       copy(properties = this.properties ++ properties)
+
+    override def withSchemaRegistryClientRetry(
+      schemaRegistryClientRetry: SchemaRegistryClientRetry[F]
+    ): AvroSettings[F] = copy(schemaRegistryClientRetry = schemaRegistryClientRetry)
 
     override def createAvroDeserializer(
       isKey: Boolean
@@ -184,7 +206,7 @@ object AvroSettings {
 
   private[this] def create[F[_]](
     schemaRegistryClient: F[SchemaRegistryClient]
-  )(implicit F: Sync[F]): AvroSettings[F] =
+  )(implicit F: Async[F]): AvroSettings[F] =
     AvroSettingsImpl(
       schemaRegistryClient = schemaRegistryClient,
       properties = Map.empty,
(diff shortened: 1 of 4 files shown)


modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSettings.scala
```diff
@@ -184,7 +206,7 @@ object AvroSettings {
 
   private[this] def create[F[_]](
     schemaRegistryClient: F[SchemaRegistryClient]
-  )(implicit F: Sync[F]): AvroSettings[F] =
+  )(implicit F: Async[F]): AvroSettings[F] =
```
gordon-rennie, line 209: We need `Async` here now because the default retry needs `Async`. If this sounds problematic (noting this is effectively the only constructor) we could provide something like a new `createNoRetry[F[_]: Sync]` - let me know what you think.

modules/vulcan/src/main/scala/fs2/kafka/vulcan/SchemaRegistryClientRetry.scala
```diff
@@ -0,0 +1,62 @@
...
+          action.handleErrorWith(
+            err =>
+              if ((attempt + 1) <= 10 && isRetriable(err))
+                F.sleep((10 * Math.pow(2, attempt.toDouble)).millis) >> retry(attempt + 1, action)
+              else
+                F.raiseError(err)
+          )
+
+        retry(attempt = 1, action)
+      }
```
gordon-rennie, line 51: The Java SchemaRegistryClient uses an underlying [`RestService`](https://github.com/confluentinc/schema-registry/blob/618e59e5a68d335cb5fa94bf7bb0777d2978aa6e/client/src/main/java/io/confluent/kafka/schemaregistry/client/rest/RestService.java#L145-L146) with 60 second timeouts. Ten attempts is probably far too many. I'm thinking of mitigating by reducing to ~3 attempts with a more aggressive backoff, 100 millis and 1000 millis.

bplommer:
This looks great! I think I'm happy with the approach - any thoughts on this @vlovgr?

gordon-rennie:
Closing as master has drifted significantly - the last time I reviewed the situation, new changes on master made this approach unattractive and it would be better implemented by having a pure interface wrapping the Java schema registry client, then using retry combinators on the call sites.

PR #730, opened 2021-10-20
Add deleteConsumerGroups method

keirlawson:


LMnet:
You need to add exclusion rule for a new method. In the build.sbt

Overl I'm +1

PR #843, opened 2022-01-24
[proposal] Externalize schema registry client

geirolz:
Recently I had to use `fs2-kafka` in prod with Protobuf. Since I didn't find any module or library ready to use I started to develop my own implementation ( delegating serialization and deserialization to the confluent library as the Avro module). 
Said that I thought that this could be useful for someone else.

I've forked the project starting a new module `fs2-kafka-protobuf` but I realized that a dedicated module for the schema-registry client could be useful as well. The next step could be creating the protobuf module that uses this new module. 

I'm not sure you want to publish this module.
In any case, let me know what do you think about this


geirolz:
@bplommer any news about this? Thanks 👍

bplommer:
Seems like a good idea - I'll try and take a look soon.

PR #844, opened 2022-01-27
Add support for using CooperativeStickyAssignor with KafkaConsumer

biochimia:
KafkaConsumer.partitionsMapStream would drop all partition streams
whenever a partition revocation was advertised, but would only create
streams for newly assigned partitions. This worked fine with the default
assignor, which revokes all existing assignments before new assignments
are issued during a rebalance operation.

The CooperativeStickyAssignor, attempts to keep assignments stable,
performing rebalance operations in two steps: on a first pass a minimal
set of assignments are revoked, these are then redistributed to balance
the consumer cluster.

In order to support this assignor, the consumer needs to keep track of
the current set of assignments and, in the case of partitionsMapStream,
recreate partition streams for partitions that were kept. While the
underlying org.apache.kafka.clients.consumer.KafkaConsumer does this for
us, this follows existing practice in KafkaConsumer.assignmentStream to
keep track of the currently assigned set.

Going forward, it would make sense to implement partitionedStream
independently of partitionsMapStream, so as to avoid recreating streams
for sticky partitions, during a rebalance operation.

New tests were added to validate that assignmentStream and
partitionsMapStream work with the CooperativeStickyAssignor. The former
was working out of the box, as the stream kept track of the assigned set
on its own. For the latter use case, a timeout was added, as revoking
assigned partitions would make the test hang indefinitely.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
+++ b/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
@@ -312,7 +312,7 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
         val records = withRebalancing.records.keySetStrict
 
         val revokedFetches = revoked intersect fetches
-        val revokedNonFetches = revoked diff fetches
+        val revokedNonFetches = revoked diff revokedFetches
 
         val withRecords = records intersect revokedFetches
         val withoutRecords = revokedFetches diff records
--- a/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
+++ b/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
@@ -614,13 +614,10 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
           assert {
             keys.size.toLong == producedTotal &&
             keys.values.sum == 236 &&
-            consumer1assignments.size == 3 &&
+            consumer1assignments.size == 1 &&
             consumer1assignments(0) == Set(0, 1, 2) &&
-            consumer1assignments(1) == Set(0, 1) &&
-            consumer1assignments(2) == Set(0, 1) &&
             consumer2assignments.size == 1 &&
-            consumer2assignments(0) == Set(2) &&
-            consumer1assignments(1) ++ consumer2assignments(0) == Set(0, 1, 2)
+            consumer2assignments(0) == Set(2)
           }
         }).unsafeRunSync()
       }
--- a/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
+++ b/modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
@@ -9,7 +9,7 @@ import cats.effect.unsafe.implicits.global
 import fs2.Stream
 import fs2.concurrent.SignallingRef
 import fs2.kafka.internal.converters.collection._
-import org.apache.kafka.clients.consumer.NoOffsetForPartitionException
+import org.apache.kafka.clients.consumer.{ConsumerConfig, CooperativeStickyAssignor, NoOffsetForPartitionException}
 import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.errors.TimeoutException
 import org.scalatest.Assertion
@@ -516,6 +516,7 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
               }
             }
             .takeWhile(_.size < 200)
+            .timeout(20.seconds)
             .compile
             .drain
             .guarantee(stopSignal.set(true))
@@ -534,6 +535,91 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
             consumer1assignments(1).size < 3 &&
             consumer2assignments.size == 1 &&
             consumer2assignments(0).size < 3 &&
+            consumer1assignments(1) ++ consumer2assignments(0) == Set(0, 1, 2)
+          }
+        }).unsafeRunSync()
+      }
+    }
+
+    it("should handle rebalance with CooperativeStickyAssignor") {
+      withTopic { topic =>
+        createCustomTopic(topic, partitions = 3)
+        val produced1 = (0 until 100).map(n => s"key-$n" -> s"value->$n")
+        val produced2 = (100 until 200).map(n => s"key-$n" -> s"value->$n")
+        val producedTotal = produced1.size.toLong + produced2.size.toLong
+
+        def startConsumer(
+          consumedQueue: Queue[IO, CommittableConsumerRecord[IO, String, String]],
+          stopSignal: SignallingRef[IO, Boolean]
+        ): IO[Fiber[IO, Throwable, Vector[Set[Int]]]] =
+          Ref[IO]
+            .of(Vector.empty[Set[Int]])
+            .flatMap { assignedPartitionsRef =>
+              KafkaConsumer
+                .stream(
+                  consumerSettings[IO]
+                    .withProperties(
+                      ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG -> classOf[CooperativeStickyAssignor].getName
+                    )
+                )
+                .subscribeTo(topic)
+                .flatMap(_.partitionsMapStream)
+                .filter(_.nonEmpty)
+                .evalMap { assignment =>
+                  assignedPartitionsRef.update(_ :+ assignment.keySet.map(_.partition())).as {
+                    Stream
+                      .emits(assignment.map {
+                        case (_, stream) =>
+                          stream.evalMap(consumedQueue.offer)
+                      }.toList)
+                      .covary[IO]
+                  }
+                }
+                .flatten
+                .parJoinUnbounded
+                .interruptWhen(stopSignal)
+                .compile
+                .drain >> assignedPartitionsRef.get
+            }
+            .start
+
+        (for {
+          stopSignal <- SignallingRef[IO, Boolean](false)
+          queue <- Queue.unbounded[IO, CommittableConsumerRecord[IO, String, String]]
+          ref <- Ref.of[IO, Map[String, Int]](Map.empty)
+          fiber1 <- startConsumer(queue, stopSignal)
+          _ <- IO.sleep(5.seconds)
+          _ <- IO(publishToKafka(topic, produced1))
+          fiber2 <- startConsumer(queue, stopSignal)
+          _ <- IO.sleep(5.seconds)
+          _ <- IO(publishToKafka(topic, produced2))
+          _ <- Stream
+            .fromQueueUnterminated(queue)
+            .evalMap { committable =>
+              ref.modify { counts =>
+                val key = committable.record.key
+                val newCounts = counts.updated(key, counts.getOrElse(key, 0) + 1)
+                (newCounts, newCounts)
+              }
+            }
+            .takeWhile(_.size < 200)
+            .timeout(20.seconds)
+            .compile
+            .drain
+            .guarantee(stopSignal.set(true))
+          consumer1assignments <- fiber1.joinWithNever
+          consumer2assignments <- fiber2.joinWithNever
+          keys <- ref.get
+        } yield {
+          assert {
+            keys.size.toLong == producedTotal &&
+            keys.values.sum == 236 &&
+            consumer1assignments.size == 3 &&
+            consumer1assignments(0) == Set(0, 1, 2) &&
+            consumer1assignments(1) == Set(0, 1) &&
+            consumer1assignments(2) == Set(0, 1) &&
+            consumer2assignments.size == 1 &&
+            consumer2assignments(0) == Set(2) &&
             consumer1assignments(1) ++ consumer2assignments(0) == Set(0, 1, 2)
           }
         }).unsafeRunSync()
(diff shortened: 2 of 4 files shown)


LMnet:
Hi @biochimia! Thank you for this contribution, it's definitely a useful addition.

Implementation and tests look good to me. But could you update scaladoc for methods in the `KafkaConsume` trait? It has some mentions about rebalance behavior and it reflects only default behavior.

bplommer:
I'll try to look at this at the weekend but also happy for this be to merged based on an approval from @LMnet

sideeffffect:
Hello guys, would it be possible to have this back-ported to 1.x series? :bow:

biochimia:
> Hi @biochimia! Thank you for this contribution, it's definitely a useful addition.
> 
> Implementation and tests look good to me. But could you update scaladoc for methods in the `KafkaConsume` trait? It has some mentions about rebalance behavior and it reflects only default behavior.

Thanks for giving this a look. I updated the documentation a bit, almost minimally, but feedback is welcome.

What are your thoughts on implementing `partitionedStream` independently of `partitionsMapStream`, to avoid closing/recreating streams for partitions that are not revoked? (I could take a stab at this in a separate PR.)

modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -540,6 +541,91 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
       }
     }
 
+    it("should handle rebalance with CooperativeStickyAssignor") {
```
biochimia, line 544: This is a copy of the `"should handle rebalance"` test, adapted to use the custom assignor, along with the resulting assert.

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -312,7 +312,7 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
         val records = withRebalancing.records.keySetStrict
 
         val revokedFetches = revoked intersect fetches
-        val revokedNonFetches = revoked diff fetches
+        val revokedNonFetches = revoked diff revokedFetches
```
biochimia, line 315: I would appreciate a comment on this change in particular.

The change makes sense to me, logically: `revokedFetches` and `revokedNonFetches` add up to `revoked`. But I did not dig enough into the implications to be sure this is the intention. For instance: in which combination of pending records and pending fetches it will make a difference, and what is the consequence?

I thought I had left this out of the PR, but while it's here we might as well discuss it.

modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -707,6 +793,65 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
       }
     }
 
+    it("should stream assignment updates to listeners when using CooperativeStickyAssignor") {
```
biochimia, line 796: This is a copy of the existing `"should stream assignment updates to listeners"` test, adapted to use the custom assignor, along with the resulting asserts.

modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -698,7 +784,7 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
             consumer1Updates(2).isEmpty &&
             consumer1Updates(3).size < 3 &&
             // Startup assignments (zero), initial assignments (< 3)
-            consumer2Updates.length == 2
+            consumer2Updates.length == 2 &&
```
biochimia, line 787: I think this affected `scalatest`'s ability to produce pretty failure messages.

modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -516,6 +516,7 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
               }
             }
             .takeWhile(_.size < 200)
+            .timeout(20.seconds)
```
biochimia, line 519: Timeout added to match test with custom assignor. This test was working without it, while the one for the custom assignor would hang indefinitely.

LMnet:
@biochimia could you please clarify for me one moment: what will a user receive after rebalancing with the `CooperativeStickyAssignor` in the `partitionsMapStream` method:
1. Streams only for the newly created partitions?
2. Or streams for all partitions?

If the answer is 1, then it means that the current returning type of `partitionsMapStream` is not perfectly fit a reality. The main motivation behind `partitionsMapStream` was to provide full information about rebalance to a library user. And now it looks like `Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]` type can't fully describe each assignment. It looks like we need something like this (ignore naming for now):
```scala
case class FullAssigmnentInfo[F[_], K, V](
  newlyAssignedPartitions: Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]],
  revokedPartitions: SortedSet[TopicPartition],
  retainedPartitions: SortedSet[TopicPartition],
)
```

If the answer is 2, we have 2 variants to deal with the streams for the retained partitions:
1. They will be recreated. For me, it looks like unnecessary work. And we will lose all benefits of the `CooperativeStickyAssignor`.
2. They will not be recreated and will be passed to the `Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]` along with the streams for created streams for newly assigned partitions. In this case, there is no way for a user to understand, which streams should be started, and which should be ignored (because they were already started earlier).

I believe that to retain backward compatibility we should pass to the `partitionsMapStream` streams only for the newly created partitions and not recreated streams for the retained partitions. And it looks like we need one more method like this (ignore naming one more time):
```scala
def streamDetailed: Stream[F, FullAssigmnentInfo[F, K, V]]
```

biochimia:
> @biochimia could you please clarify for me one moment: what will a user receive after rebalancing with the `CooperativeStickyAssignor` in the `partitionsMapStream` method:
> 
> 1. Streams only for the newly created partitions?
> 2. Or streams for all partitions?

As it stands, they'll receive 2. new streams for all currently assigned partitions.

> If the answer is 1, then it means that the current returning type of `partitionsMapStream` is not perfectly fit a reality. The main motivation behind `partitionsMapStream` was to provide full information about rebalance to a library user. And now it looks like `Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]` type can't fully describe each assignment. It looks like we need something like this (ignore naming for now):
> 
> ```scala
> case class FullAssigmnentInfo[F[_], K, V](
>   newlyAssignedPartitions: Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]],
>   revokedPartitions: SortedSet[TopicPartition],
>   retainedPartitions: SortedSet[TopicPartition],
> )
> ```
> 
> If the answer is 2, we have 2 variants to deal with the streams for the retained partitions:
> 
> 1. They will be recreated. For me, it looks like unnecessary work. And we will lose all benefits of the `CooperativeStickyAssignor`.
> 2. They will not be recreated and will be passed to the `Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]` along with the streams for created streams for newly assigned partitions. In this case, there is no way for a user to understand, which streams should be started, and which should be ignored (because they were already started earlier).
> 
> I believe that to retain backward compatibility we should pass to the `partitionsMapStream` streams only for the newly created partitions and not recreated streams for the retained partitions. And it looks like we need one more method like this (ignore naming one more time):
> 
> ```scala
> def streamDetailed: Stream[F, FullAssigmnentInfo[F, K, V]]
> ```

As you mention, one motivation behind `partitionsMapStream` is to provide full assignment information to the users. The current implementation retains this behaviour: every new element in the stream includes the current assignment, and a new stream for each assigned partition.

While this is not optimal for the `CooperativeStickyAssignor` use case, it will be the least surprising behaviour for library users. They'll be able to trust that the returned map includes the full list of assignments, and they'll be able to use `.parJoinUnbounded` to parallelise processing of all streams, without worrying about getting duplicate streams.

Finally, these changes have minimal impact on the default use case, while making the `CooperativeStickyAssignor` use case work _at all_ to begin with. Currently, using the custom assignor leads to lost partition streams on rebalance.

Without breaking user expectations for the interface, I don't think that we can make `partitionsMapStream` more optimal for the `CooperativeStickyAssignor` use case.

My proposal here is to add an implementation of `partitionedStream` that does not rely on `partitionsMapStream`. `partitionedStream` could be made to emit new streams only for newly assigned partitions, without closing/recreating streams. I think this would be optimal when using `CooperativeStickyAssignor`.

bplommer:
> Hello guys, would it be possible to have this back-ported to 1.x series? Thank you so very much :bow: 

The current policy seems to be that we accept pull requests against 1.x but don't actively seek to maintain feature parity with 2.x. Would this be easy to retweet for 1.x @biochimia?

In the longer term I think we should decide to either maintain feature parity in 1.x or stop adding any new features - any thoughts on this @LMnet @vlovgr? What do the Maven download figures for 1.x vs 2.x look like?

biochimia:
> Would this be easy to retarget for 1.x?

Looking at the files I touched (implementation and test) and the diff between the `series/1.x` and `series/2.x` branches, it looks like it should be relatively straightforward to backport the patch.

LMnet:
@biochimia  after your last comment I decided to test your branch locally. And it looks like even with the `CooperativeStickyAssignor` after each rebalance all old partition streams are closed, and for each partition (even for retained partitions) new streams are started. Is this a behavior you want to achieve?

I think we should implement support for `CooperativeStickyAssignor` differently: we should _not_ close partitions streams for retained partitions. `partitionsMapStream` should receive only newly assigned partitions and they streams. For the default assignor, nothing will be changed. But for the `CooperativeStickyAssignor` it means that old streams will continue to work without recreation.

This change means that an old convention for the `partitionsMapStream` method will be violated: it will not receive a full assignment on each rebalance in all cases. But I think this is fixable only through scaladoc change. Moreover, for the default use-case, as I said before, nothing will be changed. And we could reflect this in the documentation. Moreover, nothing will be broken in terms of binary or source compatibility. So this is a safe change I believe.

This decision also means that we will not have a method with the full assignment info after each rebalance. But I already proposed a solution for this: we need one more method with the `FullAssigmnentInfo`. It could be done in a separate pull request, I don't ask to implement it in the current branch.

> My proposal here is to add an implementation of `partitionedStream` that does not rely on `partitionsMapStream`. `partitionedStream` could be made to emit new streams only for newly assigned partitions, without closing/recreating streams. I think this would be optimal when using `CooperativeStickyAssignor`.

I don't think that this is a good idea. The current behavior of all stream methods is consistent: they all work identically. But they have a different amount of details. And that's all. I think this is an excellent thing in terms of API predictability: it doesn't matter which stream method you are using. They all have the same semantic. If different stream methods will behave differently it could be surprised for users. It may become one of these tricky minor things you have to remember to use a library.

@bplommer @vlovgr I need your opinions on this design decision because we have multiple options here.

About 1.x backport: I have nothing against it if someone will volunteer this backport. But overall the current strategy about 1.x is that maintainers don't try to maintain feature parity.

vlovgr:
@LMnet I agree with everything you said. 👍

biochimia:
> @biochimia after your last comment I decided to test your branch locally.

Thank you for taking the time to give this a test run.

> And it looks like even with the `CooperativeStickyAssignor` after each rebalance all old partition streams are closed, and for each partition (even for retained partitions) new streams are started. Is this a behavior you want to achieve?

Yes, this was the intended behaviour for my changes.

> I think we should implement support for `CooperativeStickyAssignor` differently: we should _not_ close partitions streams for retained partitions. `partitionsMapStream` should receive only newly assigned partitions and they streams. For the default assignor, nothing will be changed. But for the `CooperativeStickyAssignor` it means that old streams will continue to work without recreation.
> 
> This change means that an old convention for the `partitionsMapStream` method will be violated: it will not receive a full assignment on each rebalance in all cases. But I think this is fixable only through scaladoc change. Moreover, for the default use-case, as I said before, nothing will be changed. And we could reflect this in the documentation. And nothing will be broken in terms of binary or source compatibility. So this is a safe change I believe.

My concerns with doing this are as follows (as noted, these concerns only affect the users of `CooperativeStickyAssignor`, there are no changes to the default use case):

1. The assignment map no longer includes a full assignment.
2. Consumers of `partitionsMapStream` will not be able to distinguish an assignment revocation from a partition assignment that is kept without looking at whether the individual streams have finished.
3. Consumers of `partitionsMapStream` will not be able to rely on the behaviour that each element in `partitionsMapStream`, itself a map of streams, is fully processed and all streams closed before the next element is emitted. I don't know if this can lead to surprising behaviour in common code patterns.

> This decision also means that we will not have a method with the full assignment info after each rebalance. But I already proposed a solution for this: we need one more method with the `FullAssigmnentInfo`. It could be done in a separate pull request, I don't ask to implement it in the current branch.
> 
> > My proposal here is to add an implementation of `partitionedStream` that does not rely on `partitionsMapStream`. `partitionedStream` could be made to emit new streams only for newly assigned partitions, without closing/recreating streams. I think this would be optimal when using `CooperativeStickyAssignor`.
> 
> I don't think that this is a good idea. The current behavior of all stream methods is consistent: they all work identically. But they have a different amount of details. And that's all. I think this is an excellent thing in terms of API predictability: it doesn't matter which stream method you are using. They all have the same semantic. If different stream methods will behave differently it could be surprised for users. It may become one of these tricky minor things you have to remember to use a library.
> 
> @bplommer @vlovgr I need your opinions on this design decision because we have multiple options here.

As you note, this is a design choice for the library. I bring forward my concerns so they can be discussed (and even invalidated). If there's agreement that the best path forward is to not recreate streams at the cost of having an incomplete assignment map in `partitionsMapStream`, I'm happy to update my changes.

About impact on current use cases using the default assignor, my only concern was about an implementation detail: currently we use a single `prevAssignmentFinisherRef`, we'll need to switch to independent finishers per partition. I mention this for completeness, but I'm happy for the discussion to focus on the design choices first, and come back to the implementation later.

LMnet:
@biochimia, all your concerns are correct. It is a trade-off, of course.

> As you note, this is a design choice for the library. I bring forward my concerns so they can be discussed (and even invalidated). If there's agreement that the best path forward is to not recreate streams at the cost of having an incomplete assignment map in partitionsMapStream, I'm happy to update my changes.

I think you can go on and update your changes. I'm pretty sure @bplommer will not be against it.

> About impact on current use cases using the default assignor, my only concern was about an implementation detail: currently we use a single prevAssignmentFinisherRef, we'll need to switch to independent finishers per partition. I mention this for completeness, but I'm happy for the discussion to focus on the design choices first, and come back to the implementation later.

I think you could create a single `Ref` with all finalizers. This will not have too much performance impact.

biochimia:
I've updated the PR to reflect what we've discussed. Namely, `partitionsMapStream` now keeps a per-partition finisher, and will only close streams for revoked partitions. Assignment maps emitted by `partitionsMapStream` will only include newly assigned partitions and their streams.

LMnet (approved):
Great job, @biochimia! Thanks for your contribution!

biochimia:
I re-pushed my changes (updated hash, only), to trigger the CI workflow. There was an [issue affecting GitHub actions](https://www.githubstatus.com/incidents/vvthhf8gxt80) before that meant some triggers did not fire.

sideeffffect:
This PR's CI needs trigger from a maintainer :pray:

LMnet:
@biochimia I tried to run CI multiple times and it fails on the `fs2.kafka.KafkaConsumerSpec#stream should consume all records with subscribing for several consumers`. Could you check it, please?

biochimia:
> @biochimia I tried to run CI multiple times and it fails on the `fs2.kafka.KafkaConsumerSpec#stream should consume all records with subscribing for several consumers`. Could you check it, please?

I can give this a look, but it seems that this test is flaky and fails on a race condition:

- an earlier run of the tests succeeded for the same changes: https://github.com/fd4s/fs2-kafka/actions/runs/1805421039/attempts/2
- you can see a similar failure in an unrelated PR upgrading the `testcontainers` version: https://github.com/fd4s/fs2-kafka/actions/runs/1800805195

LMnet:
@biochimia ok, got it.

@bplommer can you take a look at this pr? Or maybe we could merge it and fix all stuff by ourselves.

biochimia:
I had a look at the failing test, and was able to reproduce the failure on top of vanilla `series/2.x` with a loop in the test and some added logging to see what's going on.

- my changes to add logging and isolate the test can be found here: https://github.com/fd4s/fs2-kafka/compare/series/2.x...biochimia:biochimia/series/race-condition
- you can find detailed logs showing one successful and one failing iteration of the test here: https://github.com/biochimia/fs2-kafka/runs/5134797071?check_suite_focus=true

When the test fails, you see the first consumer being assigned all partitions and fetching the records before the second consumer is able to join. When the second consumer joins, and assignment changes, the underlying consumer [resets offsets back to 0](https://github.com/biochimia/fs2-kafka/runs/5134797071?check_suite_focus=true#step:4:243), and records are consumed again, thus failing the test with repeated records.

The race condition then seems to be about the first consumer being able to fetch and process all records before a rebalance is triggered by the second consumer joining.

sideeffffect:
Awesome news that this got merged! :tada: @biochimia how much work do you think it would take to backport this to `series/1.x` (compatible with Cats Effect 2)?

PR #888, opened 2022-03-07
Threadsafe transactions

janstenpickle:
Addresses a bug with concurrent transactional producer access discovered in #883.

Introduces a semaphore around the transactional produce `WithProducer` implementation. This is because only one transaction may be started per producer at any one time, so produce requests must be queued.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
@@ -6,7 +6,8 @@
 
 package fs2.kafka.internal
 
-import cats.effect.{Blocker, ContextShift, Resource, Sync}
+import cats.effect.concurrent.Semaphore
+import cats.effect.{Blocker, Concurrent, ContextShift, Resource, Sync}
 import cats.implicits._
 import fs2.kafka.{KafkaByteProducer, ProducerSettings, TransactionalProducerSettings}
 import fs2.kafka.internal.syntax._
@@ -35,24 +36,28 @@ private[kafka] object WithProducer {
   def apply[F[_], K, V](
     settings: TransactionalProducerSettings[F, K, V]
   )(
-    implicit F: Sync[F],
+    implicit F: Concurrent[F],
     context: ContextShift[F]
   ): Resource[F, WithProducer[F]] =
-    blockingResource(settings.producerSettings).flatMap { blocking =>
-      Resource[F, WithProducer[F]] {
-        settings.producerSettings.createProducer.flatMap { producer =>
-          val withProducer = create(producer, blocking)
+    (blockingResource(settings.producerSettings), Resource.eval(Semaphore(1))).tupled.flatMap {
+      case (_blocking, transactionSemaphore) =>
+        Resource[F, WithProducer[F]] {
+          settings.producerSettings.createProducer.flatMap { producer =>
+            val withProducer = new WithProducer[F] {
+              override def apply[A](f: (KafkaByteProducer, Blocking[F]) => F[A]): F[A] =
+                transactionSemaphore.withPermit(f(producer, _blocking))
+            }
 
-          val initTransactions = withProducer.blocking { _.initTransactions() }
+            val initTransactions = withProducer.blocking { _.initTransactions() }
 
-          val close = withProducer.blocking {
-            _.close(settings.producerSettings.closeTimeout.asJava)
+            val close = withProducer.blocking {
+              _.close(settings.producerSettings.closeTimeout.asJava)
+            }
+
+            initTransactions.as((withProducer, close))
           }
 
-          initTransactions.as((withProducer, close))
         }
-
-      }
     }
 
   private def blockingResource[F[_]: Sync: ContextShift](
--- a/modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
@@ -6,11 +6,10 @@
 
 package fs2.kafka.internal
 
-import cats.effect.concurrent.Semaphore
-import cats.effect.{Blocker, Concurrent, ContextShift, Resource, Sync}
+import cats.effect.{ContextShift, Resource, Sync}
 import cats.implicits._
-import fs2.kafka.{KafkaByteProducer, ProducerSettings, TransactionalProducerSettings}
 import fs2.kafka.internal.syntax._
+import fs2.kafka.{KafkaByteProducer, ProducerSettings}
 
 private[kafka] sealed abstract class WithProducer[F[_]] {
   def apply[A](f: (KafkaByteProducer, Blocking[F]) => F[A]): F[A]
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/fs2/kafka/internal/WithProducer.scala
```diff
@@ -35,24 +36,28 @@ private[kafka] object WithProducer {
...
-      Resource[F, WithProducer[F]] {
-        settings.producerSettings.createProducer.flatMap { producer =>
-          val withProducer = create(producer, blocking)
+    (blockingResource(settings.producerSettings), Resource.eval(Semaphore(1))).tupled.flatMap {
+      case (_blocking, transactionSemaphore) =>
+        Resource[F, WithProducer[F]] {
+          settings.producerSettings.createProducer.flatMap { producer =>
+            val withProducer = new WithProducer[F] {
+              override def apply[A](f: (KafkaByteProducer, Blocking[F]) => F[A]): F[A] =
+                transactionSemaphore.withPermit(f(producer, _blocking))
```
janstenpickle, line 48: @bplommer Is this what you meant when you said to put the semaphore in `WithProducer`?

I don't think it's a huge problem, but this will keep the permit for potentially non-transactional operations

bplommer: Yes, that's what I meant.

Hmm - I agree it's not a huge problem, but it's also not great that `metrics` calls will also potentially block on getting a permit that they don't actually need 🤔

Maybe we should have a separate `WithTransactionalProducer` trait where `apply` has the signature `def apply[A](f: (KafkaByteProducer, Blocking[F], Resource[F, Unit]) => F[A]): F[A]` so that the caller can control how the semaphore is used - what do you think?

(As an aside, I'm not actually convinced that `WithProducer` carries its weight as an abstraction - it would make more sense to have if the underlying producer were pooled or hot-swappable, but it isn't. But as long as we do have that abstraction, I think the lifecycle of the semaphore should be managed in the same place as the lifecycle of the producer.)

modules/core/src/main/scala/fs2/kafka/internal/package.scala
```diff
@@ -0,0 +1,15 @@
...
+  type ExclusiveAccess[F[_], A] = F[A] => F[A]
+
+  private[kafka] def blockingResource[F[_]: Sync: ContextShift](
+    settings: ProducerSettings[F, _, _]
+  ): Resource[F, Blocking[F]] =
+    settings.blocker
+      .map(Resource.pure[F, Blocker])
+      .getOrElse(Blockers.producer)
+      .map(Blocking.fromBlocker[F])
+}
```
janstenpickle, line 15: @bplommer not sure how you feel about package level methods, but I put this here as it is shared by both `WithProducer` implementations.

modules/core/src/main/scala/fs2/kafka/internal/package.scala
```diff
@@ -0,0 +1,15 @@
+package fs2.kafka
+
+import cats.effect.{Blocker, ContextShift, Resource, Sync}
+
+package object internal {
+  type ExclusiveAccess[F[_], A] = F[A] => F[A]
```
janstenpickle, line 6: Rather than add another abstraction for exclusive access I made a type alias. This could cause type inference problems in the future as it's not partially applied, but I think it's probably fine...

modules/core/src/main/scala/fs2/kafka/internal/WithTransactionalProducer.scala
```diff
@@ -0,0 +1,63 @@
...
+        Resource[F, WithTransactionalProducer[F]] {
+          settings.producerSettings.createProducer.flatMap { producer =>
+            val withProducer = create(producer, blocking, accessSemaphore)
+
+            val initTransactions = withProducer.blocking { _.initTransactions() }
+
+            val close = withProducer.exclusiveAccess {
+              case (producer, blocking) =>
+                blocking(producer.close(settings.producerSettings.closeTimeout.asJava))
+            }
```
janstenpickle, line 45: I realised that closing the producer should probably have exclusive access too!

bplommer: Not sure whether it's necessary but definitely shouldn't do any harm

modules/core/src/main/scala/fs2/kafka/internal/package.scala
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package fs2.kafka
+
+import cats.effect.{Blocker, ContextShift, Resource, Sync}
+
+package object internal {
+  type ExclusiveAccess[F[_], A] = F[A] => F[A]
```
bplommer, line 12:
```suggestion
private[kafka] package object internal {
  private[kafka] type ExclusiveAccess[F[_], A] = F[A] => F[A]
```

modules/core/src/main/scala/fs2/kafka/internal/package.scala
bplommer: As long as it's all package-private let's go with it - we can always refactor it later on.

modules/core/src/main/scala/fs2/kafka/internal/package.scala
bplommer: Again, not too worried as long as it's package-private.

bplommer (approved):

modules/core/src/main/scala/fs2/kafka/internal/WithTransactionalProducer.scala
janstenpickle: Yes, I think you're right it might not be necessary, because the producer will wait for any in flight records to be flushed or the provided timeout to expire. From the perspective of not letting any other records get added to the producer during shutdown this is quite nice, but this shouldn't happen thanks to `Resource`, unless someone has done something silly like leak the producer out of the resource.

It also changes the behaviour of the `closeTimeout` setting for the transactional producer; we're using the semaphore to sequence transactions where we wait for the client buffer to flush and transaction comitted to the broker, the `closeTimeout` effectively becomes redundant when we use the semaphore with `close` because the buffer will always be empty. Given this do you think it's better to remove the `exclusiveAccess` call here?

modules/core/src/main/scala/fs2/kafka/internal/package.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright 2018-2022 OVO Energy Limited
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package fs2.kafka
+
+import cats.effect.{Blocker, ContextShift, Resource, Sync}
+
+private[kafka] package object internal {
```
bplommer, line 11:
```suggestion
package object internal {
```

janstenpickle:
@bplommer it's probably not a problem because no one has reported any issues with concurrent access, but this change wll make concurrent access to a single producer effectively single-threaded. I wonder if it's worth updating the documentation noting this, maybe even providing guidance on pooling producers?

bplommer:
@janstenpickle that does sound worthwhile - I'm happy to merge this PR without it and am unlikely to get to it myself in the near future, but if you want to add that change here or in a follow-up PR (or just open an issue) you're more than welcome to 😅

janstenpickle:
Yes, of course. Happy to add the warning, I haven't experimented with pooled transactional producers myself, so probably won't add anything other than a vague recommendation for now. As you said elsewhere pooling functionality something that could be added to `3.x`.

PR #890, opened 2022-03-08
Update transactional producer docs

janstenpickle:
Relates to #888, will need to wait for it to be ported to `series/2.x`

Adds note on concurrent access performance constraints.


bplommer:
Thanks! Just squeezing in a couple of other things, then I'll get a release train going.

janstenpickle:
@bplommer I'm happy to do the porting from `1.x` to `2.x` if you'd like?

bplommer:
Go for it!

PR #892, opened 2022-03-09
Threadsafe transactions

janstenpickle:
Introduces a new `WithTansactionalProducer` implementation. This is
because only one transaction may be started per producer at any one time,
so produce requests must be queued.

Ported from `series/1.x`

Unblocks #890


bplommer (approved):
Thanks!

PR #897, opened 2022-03-09
Use ConsumerGroupMetadata for offset commits for stronger semantics

bplommer:
As stated in https://kafka.apache.org/documentation/#upgrade_3_0_0:
> The `Producer#sendOffsetsToTransaction(Map offsets, String consumerGroupId)` method has been deprecated. Please use `Producer#sendOffsetsToTransaction(Map offsets, ConsumerGroupMetadata metadata)` instead, where the ConsumerGroupMetadata can be retrieved via `KafkaConsumer#groupMetadata()` for stronger semantics. Note that the full set of consumer group metadata is only understood by brokers or version 2.5 or higher, so you must upgrade your kafka cluster to get the stronger semantics. Otherwise, you can just pass in `new ConsumerGroupMetadata(consumerGroupId)` to work with older brokers. See [KIP-732](https://cwiki.apache.org/confluence/x/zJONCg) for more details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 target/
 .metals/
 .vscode/
+.bsp/
+.bloop/
+metals.sbt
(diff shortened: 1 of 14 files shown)


aartigao:
Ben, after so many years I think it's time to do some cleanup and close old drafts to avoid noise in the Pull Request section and focus on newer things.

OFC, feel free to reopen it if you think it's worth 👍🏽

PR #898, opened 2022-03-12
Introduce type-level distinction between key and value serializers/deserializers

bplommer:
Introduces phantom types `KeyOrValue`, `Key` and `Value`, and generalises `Serializer` to `GenSerializer` with an extra type parameter for `KeyOrValue` subtype - `Serializer`, `KeySerializer` and `ValueSerializer` become aliases with different values set for that type parameter. Same goes for `Deserializer`.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -40,12 +40,12 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
   /**
     * The `Deserializer` to use for deserializing record keys.
     */
-  def keyDeserializer: F[Deserializer[F, K]]
+  def keyDeserializer: F[KeyDeserializer[F, K]]
 
   /**
     * The `Deserializer` to use for deserializing record values.
     */
-  def valueDeserializer: F[Deserializer[F, V]]
+  def valueDeserializer: F[ValueDeserializer[F, V]]
 
   /**
     * A custom `ExecutionContext` to use for blocking Kafka operations. If not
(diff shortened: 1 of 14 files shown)


modules/core/src/main/scala/fs2/kafka/Deserializer.scala
```diff
@@ -16,7 +16,7 @@ import java.util.UUID
   * Functional composable Kafka key- and record deserializer with
   * support for effect types.
   */
-sealed abstract class Deserializer[F[_], A] {
+sealed abstract class GenDeserializer[-T <: KeyOrValue, F[_], A] {
```
LMnet, line 19: What is `Gen`? Generic?

bplommer: I'm not actually sure - I'm following the pattern of things like `GenSpawn` in cats-effect. Either generic or generalized I guess...

LMnet: Maybe `GenericDeserializer` would be better? I believe that self-descriprive naming is more important than saving 4 letters to type.

vlovgr (approved):

modules/core/src/main/scala/fs2/kafka/Deserializer.scala
bplommer: Yeah that's fair 😅

modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
```diff
@@ -40,12 +40,12 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
   /**
     * The `Deserializer` to use for deserializing record keys.
     */
-  def keyDeserializer: F[Deserializer[F, K]]
+  def keyDeserializer: F[KeyDeserializer[F, K]]
```
bplommer, line 43: We should probably switch to allocating these in `Resource` - it allows calling `close` when wrapping Java serdes, allows widening from `Deserializer` to `KeyDeserializer`/`ValueDeserializer` without needing `Functor`, and better conveys why we're wrapping things in an effect.

LMnet (approved):

PR #899, opened 2022-03-13
Add config to emit empty chunks when fetch returns nothing

bplommer:


rodrigo-molina:
Today, we have some Kafka consumers buffering data and flushing it based on either a buffer max size or a max lag latency. 
In our legacy applications we have done what's actually done in this PR: just propagate empty polls so the consumer has always a tick interval (poll interval) to check the latency.  We have achieved the same result with fs2 by merging the consumer stream with an infinite time ticks stream. Therefore, our final stream has either Kafka messages or a time ticks.

As final a thought, having the time ticks separated from the poll interval let us tweak them independently. I'm trying to think a scenario when we would prefer to bind them.

PR #901, opened 2022-03-13
Simplify ProducerRecords types by removing `passthrough`

bplommer:
The functionality for producer passthrough values seems to be completely orthogonal to the functionality for actually producing records - it can be replicated in user code by just doing `produce(...).map(_.tupleLeft(passthrough))` - this PR removes specific passthrough functionality, and as a result is able to replace the `ProducerRecords` and `TransactionalProducerRecords` classes with aliases for `Chunk`.

The only place where user code can't implement passthroughs is when using `KafkaProducer.pipe` - I've added `pipeWithPassthrough` to implement this using tuples.

Docs need updating - I'll do so if feedback on this is positive.


modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -205,25 +204,40 @@ object KafkaProducer {
...
+  def pipe[F[_], K, V](
     settings: ProducerSettings[F, K, V]
   )(
     implicit F: Async[F],
     mk: MkProducer[F]
-  ): Pipe[F, ProducerRecords[P, K, V], ProducerResult[P, K, V]] =
+  ): Pipe[F, ProducerRecords[K, V], ProducerResult[K, V]] =
     records => stream(settings).flatMap(pipe(settings, _).apply(records))
 
+  def pipeWithPassthrough[F[_], K, V, P](
```
LMnet, line 215: Actually I'm not sure that we need this.

modules/core/src/main/scala/fs2/kafka/ProducerRecords.scala
```diff
@@ -6,73 +6,20 @@
...
-
-  /** The records to produce. Can be empty for passthrough-only. */
-  def records: Chunk[ProducerRecord[K, V]]
-
-  /** The passthrough to emit once all [[records]] have been produced. */
-  def passthrough: P
-}
+import cats.Traverse
 
 object ProducerRecords {
```
LMnet, line 14: Maybe it's worth to move this companion to the type alias definition to improve discoverability? Currently type itself defined in one place and its component in another.

LMnet (changes requested):

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
bplommer: Oh, can user code be rewritten to do the same thing with `pipe`?

modules/core/src/main/scala/fs2/kafka/ProducerRecords.scala
bplommer: Sure, I don’t want to bloat the package object file too much but I guess that’s ok for just a few constructors

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
vlovgr: I agree, we shouldn't keep adding pipe combinators. We can keep `pipe` but if you need something more involved, then you can use the non-pipe combinators.

vlovgr:
This is great @bplommer! 🎉 Been wanting to do this for quite some time, since it was just done hastily in the same manner as Akka Streams in the very early days of the library.

modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
```diff
@@ -234,6 +234,7 @@ sealed abstract class ProducerSettings[F[_], K, V] {
 }
 
 object ProducerSettings {
+  val DefaultParallelism: Int = 10000
```
bplommer, line 237: I have a couple of concerns about this:
- I don't think it should be in `ProducerSettings` - it should just be a parameter to `pipe`.
- I'm not convinced it works as adverstised (controlling batch size). It limits the number of `ProduceResults` we're waiting for at any one time, but I'm not sure that's needed - the Java producer already maintains a buffer and blocks (with a default timeout of 60 seconds) when the buffer is full. Can't we just rely on that for backpressure and have unlimited parallelism on waiting for the results? (@vlovgr any thoughts on this?)

vlovgr: I think we can have unbounded parallellism instead, yes.

docs/src/main/mdoc/quick-example.md
```diff
@@ -37,12 +37,17 @@ object Main extends IOApp {
...
-        .through(KafkaProducer.pipe(producerSettings))
-        .map(_.passthrough)
-        .through(commitBatchWithin(500, 15.seconds))
+        }.through { offsetsAndProducerRecords =>
+          KafkaProducer.stream(producerSettings).flatMap { producer =>
+            offsetsAndProducerRecords.evalMap { 
+              case (offset, producerRecord) => 
+                producer.produce(producerRecord)
+                  .map(_.as(offset))
+            }.parEvalMap(Int.MaxValue)(identity)
```
bplommer, line 48: This is a bit awkward - I wonder if we could convince the fs2-kafka maintainers to add `parFlattenUnbounded` 🤔

bplommer:
Oh, I think we can do the same for `ProducerResult`!

docs/src/main/mdoc/quick-example.md
vlovgr: If not, maybe at least `parEvalMapUnbounded`?

bplommer: I opened a PR for `parEvalMapUnbounded` - https://github.com/typelevel/fs2/pull/2849.

LMnet (approved):

PR #902, opened 2022-03-13
Move deserialization from KafkaConsumerActor to KafkaConsumer

bplommer:
Postpones deserialising records until they are pushed onto a consumer stream, reducing the complexity of `KafkaConsumerActor`.

This means that a deserialization error in a partitioned stream will cause only the specific partition stream to fail, rather than crash the actor - I think that's a good thing, but maybe we should reproduce the existing behaviour (with a configurable toggle) in the 2.x series?

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -28,6 +28,7 @@ import scala.annotation.nowarn
 import scala.collection.immutable.SortedSet
 import scala.concurrent.duration.FiniteDuration
 import scala.util.matching.Regex
+import org.apache.kafka.clients.consumer.ConsumerConfig
 
 /**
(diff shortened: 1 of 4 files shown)


LMnet (approved):
Nice simplification!

bplommer:
I think I'll go ahead and merge this with the behaviour change, but put out a release candidate before the next version to give a chance to test it in the wild.

LMnet:
> This means that a deserialization error in a partitioned stream will cause only the specific partition stream to fail, rather than crash the actor - I think that's a good thing, but maybe we should reproduce the existing behaviour (with a configurable toggle) in the 2.x series?

I think we should change the behavior.

bplommer:
Meaning - you think we should release the change from this PR as-is?

LMnet:
> Meaning - you think we should release the change from this PR as-is?

Yes. I don't think that the current behavior could be considered by someone as an intended or as a "feature".

Issue #903, opened 2022-03-14
Simplification of consumer internals


LMnet:
From [here](https://github.com/fd4s/fs2-kafka/issues/889):

At the current moment, consumer internals are pretty complex: we have a `KafkaConsumer` itself, a hand-maded `KafkaConsumerActor`, a queue for messages. Also, there is some complexity with different commit methods. And most of this complexity is to make all calls to a java consumer linearized. But actually, to achieve this we could use a simple `Semaphore(1)`. Under the hood, this semaphore uses a queue (like an actor), but all low-level queue-related machinery is hidden. With this, we could dramatically reduce the complexity of library internals and make it more contributor-friendly. I think I have to create an issue about these thoughts.

bplommer:
That sounds good - the only thing I wonder is how we would handle prioritising client requests over polls? I wonder if instead of switching to `cats.effect.Semaphore` we should instead/first replace the different kind of actor requests with a single semaphore-like permit request

bplommer:
I'm thinking something like https://github.com/fd4s/fs2-kafka/pull/906 - does that fit with what you had in mind?

aartigao:
I agree with @LMnet that current code is not contributor-friendly. That's my experience at least :sweat_smile: I find it difficult to follow and I'm pretty confident that we can simplify it by removing completely `KafkaConsumerActor`.

I'm working on a POC to simplify the internals (let's see how far I can get).

About how to prioritize client requests over polls: wouldn't [Priority Queue](https://typelevel.org/cats-effect/docs/std/pqueue) be the answer to this?

PR #906, opened 2022-03-15
Use semaphore-like permits for some actor operations

bplommer:
This was supposed to be a demo but I got a bit carried away - inspired by #903

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -275,32 +275,29 @@ object KafkaConsumer {
           streamId: StreamId,
           assignmentRef: Ref[F, Map[TopicPartition, Deferred[F, Unit]]],
           partitionsMapQueue: PartitionsMapQueue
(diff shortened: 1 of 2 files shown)


bplommer:
Oops, I broke something somewhere along the line 😬

LMnet:
I checked these changes briefly. They may be good by itself, but I talked about a different approach: I wanted to completely get rid of a `KafkaConsumerActor`. We don't need it actually. We could call а java consumer from the `fs2.kafka.KafkaConsumer` through semaphore without `KafkaConsumerActor`.

To create polling we could run a fiber that uses the same semaphore and the same java consumer and run this fiber in the background.

vlovgr:
Ideally, we should process all fetch requests before polling, since partitions without fetch requests will remain paused when polling (as you note, we aren't really doing exactly that today). If it makes a difference in practice versus just processing everything FIFO is difficult to say. The only contract we really have to adhere to is to ensure poll is called at least every `max.poll.interval.ms`.

bplommer:
Is there any objection to merging these changes as a step towards using a semaphore for everything?

LMnet (approved):

PR #907, opened 2022-03-15
withParallelism default 100 -> 10000 in docs

jbwheatley:
the default value for `parallelism` in `ProducerSettings` is 10000 https://github.com/fd4s/fs2-kafka/blob/19f5210f0339f65d41cc4cb30a750c59754d78f1/modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala#L326

Just fixing the docs to match this.

### DIFF
--- a/docs/src/main/mdoc/producers.md
+++ b/docs/src/main/mdoc/producers.md
@@ -130,7 +130,7 @@ The following settings are specific to the library.
 
 - `withCloseTimeout` controls the timeout when waiting for producer shutdown. Default is 60 seconds.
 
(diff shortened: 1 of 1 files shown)


bplommer:
Thanks!

bplommer:
Thanks! Be aware that this setting doesn't really work the way it's advertised (see https://github.com/fd4s/fs2-kafka/pull/901#discussion_r825720073) - it's removed on the 3.x branch and we should probably deprecate it on the 2.x branch.

PR #918, opened 2022-03-27
Use permit for fetch, rework permits

bplommer:


modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -133,17 +135,17 @@ object KafkaConsumer {
...
         type PartitionsMap = Map[TopicPartition, Stream[F, CommittableConsumerRecord[F, K, V]]]
         type PartitionsMapQueue = Queue[F, Option[PartitionsMap]]
 
-        def createPartitionStream(
+        def partitionStream(
           streamId: StreamId,
           partition: TopicPartition,
           assignmentRevoked: F[Unit]
-        ): F[Stream[F, CommittableConsumerRecord[F, K, V]]] =
+        ): Stream[F, CommittableConsumerRecord[F, K, V]] = Stream.force {
```
bplommer, line 148: This signature change means each partition stream will be initialized when the user starts pulling from it, rather than before the map is emitted - I don't see why that should be a problem, but if it is then this change can easily be reversed.

vlovgr (approved):

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -178,13 +179,34 @@ object KafkaConsumer {
...
+                deferred.complete(_).void
+
+              val fetch: F[PartitionResult] = withPermit {
+                val assigned =
+                  withConsumer.blocking {
+                    _.assignment.contains(partition)
+                  }
+
+                def storeFetch: F[Unit] =
+                  actor.ref.modify { state =>
```
LMnet, line 193: Looks a bit weird that we need to modify internal actor state from the another component.

bplommer: Yeah, I guess it's an artifact of an iterative transition away from having the actor be a separate encapsulated component. Maybe we should reconceptualise the state as "consumer state" rather than "actor state".

LMnet: Up to you. I mentioned it only to make sure that this is intentional.

PR #923, opened 2022-03-28
Update testcontainers version, set image name and version explicitly

bplommer:
Also adds support for running tests on Arm64

### DIFF
--- a/modules/core/src/test/resources/logback-test.xml
+++ b/modules/core/src/test/resources/logback-test.xml
@@ -6,7 +6,7 @@
     </encoder>
   </appender>
 
-  <root level="ERROR" additivity="false">
+  <root level="INFO" additivity="false">
     <appender-ref ref="STDOUT"/>
   </root>
 </configuration>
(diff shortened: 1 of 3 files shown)


modules/core/src/test/resources/logback-test.xml
```diff
@@ -6,7 +6,7 @@
     </encoder>
   </appender>
 
-  <root level="ERROR" additivity="false">
+  <root level="INFO" additivity="false">
```
bplommer, line 9: This makes logs a lot noisier but I think it's more likely to be what you want if you're actually trying to track down a problem in them.

modules/core/src/test/scala/fs2/kafka/BaseKafkaSpec.scala
```diff
@@ -66,7 +66,15 @@ abstract class BaseKafkaSpec extends BaseAsyncSpec with ForEachTestContainer {
 
   override def runTest(testName: String, args: Args) = super.runTest(testName, args)
 
-  override val container: KafkaContainer = new KafkaContainer()
+  private val imageVersion = "7.0.1"
```
bplommer, line 69: Ideally we'd test against both the earliest and latest supported version

LMnet (approved):

Issue #486, opened 2021-01-18
Nondeterministic test failure - duplicate consumption of records


bplommer:
https://github.com/fd4s/fs2-kafka/pull/484/checks?check_run_id=1721845895#step:6:600

should consume all records with subscribing for several consumers *** FAILED *** (11 seconds, 272 milliseconds)
Vector((key-1,value->1), (key-4,value->4), (key-0,value->0), (key-2,value->2), (key-3,value->3), (key-1,value->1), (key-4,value->4), (key-0,value->0), (key-2,value->2), (key-3,value->3)) did not contain the same elements as Vector((key-0,value->0), (key-1,value->1), (key-2,value->2), (key-3,value->3), (key-4,value->4)) (KafkaConsumerSpec.scala:88)

bplommer:
Happened again in https://github.com/fd4s/fs2-kafka/runs/1762306979

bplommer:
And https://github.com/fd4s/fs2-kafka/pull/503/checks?check_run_id=1771040828

kalanzai:
@bplommer We are seeing issues where our kafka consumers using `fs2-kafka` version `1.8.0` where they consume records multiple times. Did you ever find out what happened with this test?

LMnet:
Just for clarification, is it about `fs2.kafka.KafkaConsumerSpec#stream should consume all records with subscribing for several consumers`? If yes, we have some extra info about it from @biochimia: [1](https://github.com/fd4s/fs2-kafka/pull/844#issuecomment-1033562627), [2](https://github.com/fd4s/fs2-kafka/pull/844#issuecomment-1034437836)

biochimia:
This is what I found:

- the test has a race condition in that it is not able to ensure both consumers have joined the consumer group when they start consuming records;
- a consumer that is able to join the consumer group alone will typically consume all records;
- because consumer offsets are not stored, when the second consumer joins the group and a rebalance occurs, all records will be consumed again—thus failing the test.

Outside the test case, this situation would be avoided by ensuring offsets are committed regularly. That said duplicate processing of events is part of reading from Kafka with multiple consumers, specially around rebalance operations. Assumedly, it's possible to implement exactly once semantics with some costs.

For the test itself, I think we'd have to ensure `poll` is called without consuming the records. That should allow both consumers to join the group, and perhaps test any race conditions in the code that discards fetched records during a rebalance.

I played with the test a bit but haven't found a way to do this, so far.

[...]

PR #924, opened 2022-03-29
Use foralltestcontainer

bplommer:


bplommer:
With this change, the occasional failure in #486 now happens consistently - this makes me strongly think the issue is with the production code and not the test, and that we should change the test to not expect exactly-once delivery.

PR #925, opened 2022-03-29
Use a shared test container for each suite

bplommer:
Reduces CI time from ~16m to ~9m. Also opens the way for further huge speedup by paralellizing with [weaver-test](https://disneystreaming.github.io/weaver-test/).

### DIFF
--- a/modules/core/src/test/scala/fs2/kafka/KafkaAdminClientSpec.scala
+++ b/modules/core/src/test/scala/fs2/kafka/KafkaAdminClientSpec.scala
@@ -180,27 +180,18 @@ final class KafkaAdminClientSpec extends BaseKafkaSpec {
           .resource[IO](adminClientSettings)
           .use { adminClient =>
             for {
-              topicNames <- adminClient.listTopics.names.map(
-                _.filterNot(_.startsWith("__confluent"))
-              )
-              _ <- IO(assert(topicNames.size == 1))
-              topicListings <- adminClient.listTopics.listings.map(
-                _.filterNot(_.name.startsWith("__confluent"))
-              )
-              _ <- IO(assert(topicListings.size == 1))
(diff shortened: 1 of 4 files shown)


modules/core/src/test/scala/fs2/kafka/KafkaConsumerSpec.scala
```diff
@@ -66,7 +66,7 @@ final class KafkaConsumerSpec extends BaseKafkaSpec {
       }
     }
 
-    it("should consume all records with subscribing for several consumers") {
+    it("should consume all records at least once with subscribing for several consumers") {
```
bplommer, line 69: This was flaking before (#486) but changing to foralltestcontainer made it fail consistently. I think we should change the test to not expect exactly-once delivery.

modules/core/src/test/scala/fs2/kafka/KafkaAdminClientSpec.scala
```diff
@@ -180,27 +180,18 @@ final class KafkaAdminClientSpec extends BaseKafkaSpec {
...
-              topicListings <- adminClient.listTopics.listings.map(
-                _.filterNot(_.name.startsWith("__confluent"))
-              )
-              _ <- IO(assert(topicListings.size == 1))
-              topicNamesToListings <- adminClient.listTopics.namesToListings.map(_.filter {
-                case (key, _) => !key.startsWith("__confluent")
-              })
-              _ <- IO(assert(topicNamesToListings.size == 1))
+              topicNames <- adminClient.listTopics.names
+              topicCount = topicNames.size
```
bplommer, line 184: Instead of having a fixed expectation of how many topics there should be, we check how many there are at the start and then generate other expectations relative to that.

vlovgr (approved):

Issue #506, opened 2021-01-28
Reimplement schema registry client using http4s


bplommer:
For various reasons, the Java schema registry client has been causing classpath-related issues at work. It also has terrible diagnostics. The schema registry rest API is fairly straightforward so it's tempting to write our own implementation.

SystemFw:
On top of that, the simply cache it uses is full of `synchronized`, so we've been thinking of reimplementing it as well

Wosin:
Hey @bplommer, but the idea is to still imlement  `io.confluent.kafka.schemaregistry.client.SchemaRegistryClient` inteface right ?

Wosin:
One more question, because I think I am missing something here :) We want to actually implement our own deserialization of Avro here? Since `KafkaAvroDeserializer/Serializer` are part of confluent libraries, so to completely get rid of that we would have to reimplement deserialization.

bplommer:
> One more question, because I think I am missing something here :) We want to actually implement our own deserialization of Avro here? Since `KafkaAvroDeserializer/Serializer` are part of confluent libraries, so to completely get rid of that we would have to reimplement deserialization.

That would be a much bigger task - it's not what I had in mind here (though I have a WIP PR for it at https://github.com/fd4s/vulcan/pull/289). 

What I'm thinking is to use the regular Apache Avro serializers/deserializers that are used in e.g. `vulcan.Codec.fromBinary` and then we just re-implement the part of the binary protocol that's specific to schema registry - I guess we'd do that using either the ByteBuffer API or [scodec-bits](https://github.com/scodec/scodec-bits)

keirlawson:
Was taking a look at implementing this and was wondering, how feature-complete would such a client need to be to be worth merging? ie would you be looking got jsonschema/protobuf support alongside avro per the kafka one?

bplommer:
Hi @keirlawson sorry for never replying to this! I think just having Avro support would be enough to be worth merging, though supporting json and protobuf too would also be good of course.

kevinmeredith:
Hi @bplommer - the scope of this work is to provide a pure FP interface and implementation (using http4s) for both `io.confluent.kafka.schemaregistry.client.SchemaRegistryClient` and `io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient`?

PR #932, opened 2022-03-31
Fix auto-registration of Avro union (etc) types

### LINKED ISSUE #689: Avro union type issue with Schema Registry
Hey there,

I accidentally posted [this issue](https://github.com/fd4s/vulcan/issues/381) in the vulcan project instead of this one so I'm cross-posting the issue here for greater visibility. The gist of it is that you cannot use sealed traits (union types) as top-level subjects when publishing to Kafka and I think this is because of Confluent's underlying serialization mechanism not having proper support for union types. I managed to work around this by using a record containing a schema but I was wondering if this is something we could improve upon since Confluent's Schema Registry does seem to correctly support union types but I believe its the Java Serializer from Confluent that has this limitation.

Thank you so much for all your hard work and this amazing library

bplommer:
Resolves #689 and a major motivation for #506. Overrides the entrypoint to `KafkaAvroSerializer` to use the actual writer schema rather than guessing it from the intermediate Java representation.

### DIFF
--- a/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSettings.scala
+++ b/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSettings.scala
@@ -11,6 +11,7 @@ import cats.syntax.all._
 import fs2.kafka.internal.converters.collection._
 import fs2.kafka.internal.syntax._
 import io.confluent.kafka.schemaregistry.avro.AvroSchema
+import org.apache.avro.Schema
 import vulcan.Codec
 
 /**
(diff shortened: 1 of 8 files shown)


modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroSettings.scala
```diff
@@ -103,7 +104,16 @@ sealed abstract class AvroSettings[F[_]] {
...
     * value is being serialized.
     */
-  def createAvroSerializer(isKey: Boolean): F[(KafkaAvroSerializer, SchemaRegistryClient)]
+  def createAvroSerializer(
+    isKey: Boolean,
+    writerSchema: Option[Schema]
+  ): F[(KafkaAvroSerializer, SchemaRegistryClient)]
+
+  @deprecated("use the overload that takes an optional writer schema", "2.5.0-M3")
+  final def createAvroSerializer(
```
bplommer, line 113: I think before 3.0 we should refactor some of this to expose fewer internals.

vlovgr (approved):

PR #978, opened 2022-05-18
Optimise {Committable,}ProducerRecords `apply` methods

bastewart:
**Explanation:**

The `apply` methods on both `CommittableProducerRecords` and
`ProducerRecords` hand-roll an iteration through a `Foldable` in order
to construct an `ArrayBuffer` (initialised using the `size` of the
`Foldable`). This `ArrayBuffer` is then converted to a Scala `Array` and
a `Chunk` instantiated from this.

In the worst-case (I think) this involves 3 iterations through the
`Foldable`:
 - `size`
 - `foldLeft` to construct the `ArrayBuffer`
 - Conversion of `ArrayBuffer` to an `Array`

I think the only step of the current implementation that can be
optimised is the `size` call. For example (I think) `Vector[F].size`
is constant-time (no iteration).

The implementation has been simplified to use `Chunk#iterable` and
`Foldable[F].toIterable` instead. In the best-case this should involve
no iteration as both `toIterable` and `iterable` are optimised (e.g. if
`F` is a `Vector`) . In the worst-case this should involve 1 iteration
(construction of an intermediate `LazyList`).


bastewart:
Ah, I have just realised the tight constraint can't be fixed (easily) due to binary compatibility...

This isn't something I'm particularly familiar with (I'm not familiar at all really!), but is there any hack that can be done to lower the constraint?

bplommer:
Yep I was about to say the same thing! In fact `ProducerRecords` is just an alias for `Chunk` in 3.x, but this will still be relevant for `CommittableProducerRecords`.

bastewart:
> Yep I was about to say the same thing! In fact `ProducerRecords` is just an alias for `Chunk` in 3.x, but this will still be relevant for `CommittableProducerRecords`.

Ah nice, thanks!

Assuming this change gets merged would you prefer a new PR against the `3.x` branch, or a merge to get this transferred across?

bplommer:
Looks good, sorry for the long silence on this!

> Assuming this change gets merged would you prefer a new PR against the 3.x branch, or a merge to get this transferred across?

I'll merge it across.

bplommer (approved):

PR #979, opened 2022-05-18
Add `chunk` constructors to `{Committable,}ProducerRecords`

bastewart:
Currently, there is no way to directly construct
`CommittableProducerRecords` or `ProducerRecords` from a
`Chunk[ProducerRecord]` directly. Instead, the `apply` methods must be
used which take some `G[ProducerRecord]` where `G` has a `Foldable`
instance. Ultimately both are backed by a `Chunk` though.

Since the previous PR (#978) `apply` methods use `Chunk.iterable`
and `Foldable[G].toIterable` to construct a `Chunk`*. This process is
not optimised if the collection (`G`) is a `Chunk` and it involves
iterating though the collection. This is avoided with the new public
constructor from a `Chunk`.

*Before #978 the `apply` method would have  iterated at least twice
through the collection if supplied with a `Chunk`.


bastewart:
Sorry, this includes the changes in #978, but I can't make this target my own fork as a base branch (naturally!). If #978 is merged then the diff on this should be cleaner

Issue #992, opened 2022-06-27
A NPE is thrown from AvroDeserializer if the value is empty


channingwalton:
If a message contains no key / value then a NPE is thrown from AvroDeserializer. 
```
java.lang.NullPointerException: null
	at java.base/java.nio.ByteBuffer.wrap(ByteBuffer.java:422)
	at fs2.kafka.vulcan.AvroDeserializer$.$anonfun$using$4(AvroDeserializer.scala:30)
	at defer @ fs2.kafka.vulcan.AvroDeserializer$.$anonfun$using$3(AvroDeserializer.scala:28)
	at get @ fs2.internal.InterruptContext.$anonfun$eval$1(InterruptContext.scala:91)
	at mapN @ fs2.concurrent.Channel$.bounded(Channel.scala:144)
	at map @ fs2.kafka.ConsumerRecord$.fromJava(ConsumerRecord.scala:184)
	at map @ fs2.kafka.internal.KafkaConsumerActor.$anonfun$records$2(KafkaConsumerActor.scala:406)
	at defer @ fs2.kafka.vulcan.AvroDeserializer$.$anonfun$using$3(AvroDeserializer.scala:28)
	at get @ fs2.internal.InterruptContext.$anonfun$eval$1(InterruptContext.scala:91)
	at defer @ fs2.kafka.vulcan.AvroDeserializer$.$anonfun$using$3(AvroDeserializer.scala:28)
	at get @ fs2.internal.InterruptContext.$anonfun$eval$1(InterruptContext.scala:91)
	at mapN @ fs2.concurrent.Channel$.bounded(Channel.scala:144)
	at map @ fs2.kafka.ConsumerRecord$.fromJava(ConsumerRecord.scala:184)
	at map @ fs2.kafka.internal.KafkaConsumerActor.$anonfun$records$2(KafkaConsumerActor.scala:406)
	at traverse @ fs2.kafka.KafkaConsumer$$anon$1.$anonfun$partitionsMapStream$23(KafkaConsumer.scala:259
```

[PR](https://github.com/fd4s/fs2-kafka/pull/993)

channingwalton:
I think that the byte array [here](https://github.com/fd4s/fs2-kafka/blob/85db114efdb56fa8302a71b68ae2a6be56be57e9/modules/vulcan/src/main/scala/fs2/kafka/vulcan/AvroDeserializer.scala#L27) is null.

OneCricketeer:
[...]

PR #993, opened 2022-06-27
Fix for #992 A NPE is thrown from AvroDeserializer if the value is empty

channingwalton:
I humbly submit a tiny PR for issue #992.


bplommer:
Thanks!

Issue #994, opened 2022-06-29
Sample graceful shutdown code doesn't work with Cats Effect 3


fqshopify:
AFAICT the example shared on the [FS2 website](https://fd4s.github.io/fs2-kafka/docs/consumers#graceful-shutdown) only works with FS2 versions that are based on **Cats Effect 2** (and not with FS2 versions that are based on **Cats Effect 3**). 

# Example

Here is a [repo](https://github.com/fqshopify/fs2Graceful) with 2 commits on the `main` branch: 

1. [Cats Effect 2](https://github.com/fqshopify/fs2Graceful/commit/fcf3c33f6a02ae7842fdec11cc9705349dc34c17) commit
  When you run the unit test on this branch you will see the following expected behaviour: 
```
Starting processing of record at offset = 0
Finished processing of record at offset = 0
Starting processing of record at offset = 1
Finished processing of record at offset = 1
Starting processing of record at offset = 2
Finished processing of record at offset = 2
Starting processing of record at offset = 3
Cancelling task!
Starting graceful shutdown
Finished processing of record at offset = 3
Starting processing of record at offset = 4
Finished processing of record at offset = 4
Starting processing of record at offset = 5
Finished processing of record at offset = 5
Succeeded in graceful shutdown
```
As you can see, the app finishes processing the current record at offset = 3 and any records that were buffered internally (i.e. records at offset 4 & 5) before shutting down. Note there are more records past offset 5 that are never processed, as expected. 

2. [Cats Effect 3](https://github.com/fqshopify/fs2Graceful/commit/91f06264e99b394a2ea381e0aac105972f962c74) commit

When you run the same unit test with Cats Effect 3 versions of FS2, you see the following: 

```
Starting processing of record at offset = 0
Finished processing of record at offset = 0
Starting processing of record at offset = 1
Finished processing of record at offset = 1
Starting processing of record at offset = 2
Finished processing of record at offset = 2
Starting processing of record at offset = 3
Cancelling task!
Finished processing of record at offset = 3
Starting processing of record at offset = 4
Finished processing of record at offset = 4
Starting processing of record at offset = 5
Finished processing of record at offset = 5
Starting processing of record at offset = 6
Finished processing of record at offset = 6
Starting processing of record at offset = 7
Finished processing of record at offset = 7
Starting processing of record at offset = 8
Finished processing of record at offset = 8
Starting processing of record at offset = 9
Finished processing of record at offset = 9
Starting processing of record at offset = 10
[...]

PR #1009, opened 2022-07-07
Added docs for graceful shutdown for cats-effect 3

LMnet:
Fix for #994


bplommer:
This looks good but I want to look more closely before approving it.

Re mentioning multiple versions - if/when we migrate the docs to `sbt-typelevel-site` we can split that out into versioned documentation, which it supports via Laika.

LMnet:
> if/when we migrate the docs to sbt-typelevel-site we can split that out into versioned documentation, which it supports via Laika.

It would be great!

altrack:
I have tested this snippet with the series 3 and can confirm it doesn't lock.
CE is 3.4.9 FS2 is 3.6.1
Those are the version come with FS2-kafka 3.0.1
Got the following output after sending a SIGTERM to the running app:
```
Starting graceful shutdown
Finished stream
Succeeded in graceful shutdown
```

aartigao (approved):
Given that it was tested and works as expected, I'll be merging it because it's valuable information for the users and most of them are on CE 3.

PR #1017, opened 2022-07-13
Fix cancellation behaviour in KafkaAdminClient

### LINKED ISSUE #986: Random cancellation of calls in KafkaAdminClient
Hi, I found a bug in the implementation of `cancelToken` (https://github.com/fd4s/fs2-kafka/blob/series/2.x/modules/core/src/main/scala/fs2/kafka/internal/syntax.scala#L198). It breaks the contract of `Async.async` cancellation token. What the current implementation is actually doing is immediately executing cancelation and creating a token to cancel this cancelation. Therefore,  cancellation competes with the completion of the future in https://github.com/fd4s/fs2-kafka/blob/series/2.x/modules/core/src/main/scala/fs2/kafka/internal/syntax.scala#L215.  It causes the cancellation of requests to the admin client without any reason and random behavior. 

Here is example that illustrates the issue and proposed fixed version:

```scala
import cats.effect.syntax.all._
import cats.syntax.all._
import cats.effect._
import org.apache.kafka.common.KafkaFuture
import org.apache.kafka.common.internals.KafkaFutureImpl

import scala.concurrent.duration._

object TestFs2 extends IOApp {

  implicit final class KafkaFutureSyntax[A](
      private val future: KafkaFuture[A]
  ) extends AnyVal {

    def cancelToken[F[_]](implicit F: Async[F]): F[Option[F[Unit]]] =
      F.blocking { future.cancel(true); () }.start.map(_.cancel.some)

    def cancelTokenFixed[F[_]](implicit F: Async[F]): F[Option[F[Unit]]] =
      F.delay(F.blocking { future.cancel(true); () }.start.void.some)

  }

  override def run(args: List[String]): IO[ExitCode] = {
    for {
      _ <- IO.println("Testing original fs2 cancelToken")
      f  = new KafkaFutureImpl[String]()
      _ <- f.cancelToken[IO] // create cancel token, but never call it
      _ <- IO.sleep(1.seconds)
      _ <- IO.println(f.isCancelled.toString) // true?? bug is here, future is canceled without calling cancel token

      _     <- IO.println("Testing fixed cancelToken")
      f      = new KafkaFutureImpl[String]()
      token <- f.cancelTokenFixed[IO] // create cancel token, but do not call it yet
      _     <- IO.sleep(1.seconds)
      _     <- IO.println(f.isCancelled.toString) // false, as expected
      _     <- token.getOrElse(IO.unit) // call cancellation token (if we have it)
      _     <- IO.sleep(1.seconds)
      _     <- IO.println(f.isCancelled.toString) // true, as expected

    } yield ExitCode.Success
  }

}

```

Can you please fix it and release the version without this bug. Thanks!

bplommer:
The main change is in `syntax.scala`. Resolves #986. Additionally, when fixing that I found some other problems:

- The new `KafkaFuture` instances created by the `map` and `void` syntax don't propagate cancellation back to the original future. Therefore I've removed those ops and refactored the use sites to map on `F` instead.
- `future.cancelable.start.cancel` won't cancel the future if cancellation occurs before the `async` invocation in `cancelable` is executed. I've changed this so future creation is suspended, meaning early cancelation will prevent it from being created.


modules/core/src/test/scala/fs2/kafka/BaseKafkaSpec.scala
```diff
@@ -67,15 +64,11 @@ abstract class BaseKafkaSpec extends BaseAsyncSpec with ForAllTestContainer {
...
 
-  private lazy val imageName = Option(System.getProperty("os.arch")) match {
-    case Some("aarch64") =>
-      "niciqy/cp-kafka-arm64" // no official docker image for ARM is available yet
-    case _ => "confluentinc/cp-kafka"
-  }
+  private lazy val imageName = "confluentinc/cp-kafka"
 
-  override val container: KafkaContainer = new KafkaContainer()
+  override val container: KafkaContainer = new KafkaContainer(DockerImageName.parse(s"$imageName:$imageVersion"))
```
bplommer, line 71: Unrelated but was needed for testing locally.

bplommer:
Passed locally but failing in CI 😑

modules/core/src/main/scala/fs2/kafka/internal/syntax.scala
```diff
@@ -186,34 +188,23 @@ private[kafka] object syntax {
   implicit final class KafkaFutureSyntax[A](
     private val future: KafkaFuture[A]
```
LMnet, line 189: Just for clarification: is `KafkaFuture` eager or lazy? If it's eager, so how `kafkaFuture.cancelable[F]` affects referential transparency? Maybe we should use the same approach is for scala futures?

bplommer: Sorry, just saw this - it's eager. I've pushed a change so that creation is suspended in a functional effect - is that what you meant?

bplommer: Short answer: before my change, `foo.cancelable[F]` wasn't referentially transparent. Fortunately this is an internal API only used in one place, so easy to change.

modules/core/src/test/scala/fs2/kafka/internal/SyntaxSpec.scala
```diff
@@ -50,4 +56,32 @@ final class SyntaxSpec extends BaseSpec {
       }
     }
   }
+
+  describe("KafkaFuture.cancelable") {
+
+    it("should cancel future when fiber is cancelled (2)") {
```
bplommer, line 62:
```suggestion
    it("should cancel future when fiber is cancelled") {
```

LMnet:
:+1:

LMnet (approved):

vlovgr (approved):

PR #1020, opened 2022-07-16
Merge 2.x into 3.x

bplommer:


bplommer:
To investigate: future cancelation test introduced in #1017 is failing.

PR #1053, opened 2022-08-18
Postpone removing records until assignment is made.

### LINKED ISSUE #1051: Records are missing when second rebalance happens.
Hey,
during deployments we have noticed a strange issue with records consumption.

### Versions
FS2-Kafka Version = 3.0.0-M7 (observed also for M4)
Kafka client version = 3.2.0

### Background
Let's assume that we are working with topic called `topic`, which has `100` partitions.
Our consumer is running as a single pod (`pod-1`). It is consuming all 100 partitions. During Kubernetes rolling deployments another instance is created, let's call it `pod-2`. First rebalance is triggered, making `pod-1` and `pod-2` consuming 50 partitions each.
Let's assume that:
`pod-1` is consuming partitions 0, 1, 2, 3, ...., 49
`pod-2` is consuming partitions 50, 51, 52, 53, ...., 99
After few seconds, `pod-1` gets shutdown and `pod-2` is the only working pod. Second rebalance gets triggered and now `pod-2` is consuming all partitions.
What we have noticed is that for partitions, which were previously assigned, some messages are not consumed just after second re-assignment.
More examples below.

### Code
This issue is really hard to reproduce. It happens very rarely. 

```scala
val settings =
  ConsumerSettings[F, Option[String], Array[Byte]](
    keyDeserializer = Deserializer[F, Option[String]],
    valueDeserializer = Deserializer[F, Array[Byte]],
  ).withBootstrapServers("<configuration>")
   .withGroupId("TestGroup")
   .withAutoOffsetReset("earliest")
   .withClientId("TestClientId")
   .withEnableAutoCommit(false)
   .withPollInterval(50 millis)
  //Important note: following assignment strategies are included:
 //partition.assignment.strategy = [class org.apache.kafka.clients.consumer.RangeAssignor, class org.apache.kafka.clients.consumer.CooperativeStickyAssignor]

def recordProcessor(
  records: Stream[F, CommittableConsumerRecord[F, Option[String], Array[Byte]]]
): Stream[F, CommittableOffset[F]] =
  records.evalMap { record =>
    val recordFormatted =
      s"Published record: ${record.record.topic}-${record.record.partition}:${record.record.offset}"
    Sync[F].delay(println(recordFormatted)) >> Sync[F].pure(record.offset)
  }

KafkaConsumer
  .stream(settings)
  .evalTap(_.subscribeTo("topic"))
  .flatMap(_.partitionedStream.map(recordProcessor))
  .map(_
    .through(
      _.groupWithin(100, 50 millis)
        .evalMap(CommittableOffsetBatch.fromFoldable(_).commit)
    )
  )
  .parJoinUnbounded
  .scope
```
### Example issued output

```
//pod-2 has been started. Consumer is being prepared...
INFO o.a.k.c.c.i.ConsumerCoordinator [Consumer clientId=TestClientId, groupId=TestGroup] Setting offset for partition topic-59 to the committed offset FetchPosition{offset=1000028, offsetEpoch=Optional.empty, ...}
//pod-1 has been shutdown. Rebalance has started again and now all partitions are assigned to pod-2
INFO o.a.k.c.c.i.ConsumerCoordinator [Consumer clientId=TestClientId, groupId=TestGroup] Setting offset for partition topic-59 to the committed offset FetchPosition{offset=1000032, offsetEpoch=Optional.empty, ...}
Published record: topic-59:1000037 //Offset is not moved back to 1000032, but it should, since Range (EAGER) 
// Commit happens, setting offset to 1000038.
Published record: topic-59:1000040 //Missed records: 1000038, 1000039, which exist in the topic but were skipped by consumer.
Published record: topic-59:1000041
Published record: topic-59:1000042
```

MrKustra94:
Closes #1051


modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -483,6 +489,7 @@ private[kafka] object KafkaConsumerActor {
   type StreamId = Int
 
   final case class State[F[_], K, V](
+    assignments: Set[TopicPartition],
```
MrKustra94, line 492: I did it this way because I experienced deadlock when I tried to do so with `withConsumer(_.assignments().toSet)`

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -135,13 +135,36 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
+          state
+            .withRebalancing(false)
+            .withAssignments(assigned)
+        )
       }
+      updatedState <- ref.get
+      _ <- action >>
+        log(AssignedPartitions(assigned, updatedState)) >>
+        updatedState.onRebalances.foldLeft(F.unit)(_ >> _.onAssigned(assigned))
+    } yield ()
```
bplommer, line 166: Instead of a separate `ref.get` can we return the new state from `ref.modify`:
```suggestion
    ref
      .modify { state =>
        removeRevokedRecords
          .run(
            state
              .withRebalancing(false)
              .withAssignments(assigned)
          )
          .map(stateAndAction => stateAndAction._1 -> stateAndAction)
      }
      .flatMap {
        case (updatedState, action) =>
          action >> log(AssignedPartitions(assigned, updatedState)) >>
            updatedState.onRebalances.foldLeft(F.unit)(_ >> _.onAssigned(assigned))
      }
```

MrKustra94: Yeah, I didn't notice it could be done this way.
Great thanks! 👍

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -135,13 +135,39 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
     res.start.void
   }
 
-  private[this] def assigned(assigned: SortedSet[TopicPartition]): F[Unit] =
+  private[this] def assigned(assigned: SortedSet[TopicPartition]): F[Unit] = {
+    def withState[A] = StateT.apply[Id, State[F, K, V], A](_)
+
+    val removeRevokedRecords = withState { st =>
```
lolczak, line 141: Based on the documentation, the `assigned` parameter only includes newly added partitions, so there is no need to revoke anything. It appears to be an unnecessary sanity check.

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -175,44 +201,29 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
...
         val revokedFetches = revoked intersect fetches
-        val revokedNonFetches = revoked diff revokedFetches
 
         val withRecords = records intersect revokedFetches
         val withoutRecords = revokedFetches diff records
 
         (for {
           completeWithRecords <- completeWithRecords(withRecords)
           completeWithoutRecords <- completeWithoutRecords(withoutRecords)
-          removeRevokedRecords <- removeRevokedRecords(revokedNonFetches)
```
lolczak, line 208: Why don't you clean up the `record` container from records that belong to revoked partitions and have no downstream sinks?

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -276,7 +286,7 @@ private[kafka] final class KafkaConsumerActor[F[_], K, V](
     def pollConsumer(state: State[F, K, V]): F[ConsumerRecords] =
       withConsumer
         .blocking { consumer =>
-          val assigned = consumer.assignment.toSet
+          val assigned = state.assignments
```
lolczak, line 289: Could you please explain how your changes address the issue?  The current codebase seems it doesn't have any flaw with managing assigned partitions.

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
MrKustra94: Firstly RebalanceListener calls a method to revoke partitions. Previously in this step revoked partitions had records removed. The problem here was lying in combination of Greedy rebalance protocols + paused partitions. Strangely, when paused partitions are cleared, fetching new records does not start from latest committed offset. It started from saved locally position. It caused the following problem: paused partitions records were revoked here, and when fetching new records started, this started from latest removed record offset + 1. This issue fortunately could be mitigated with Cooperative protocol, which on RebalanceListener methods used only revoked and only newly assigned partitions as opposed to Greedy protocol - all partitions were revoked and all partitions were assigned, discarding local state pre-rebalance.

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
MrKustra94: Described in above comment.
The whole fix is based on postponing removing records - we wait until assignment to make sure that we can revoke records. This is a small step handling Greedy rebalance protocols.

modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
MrKustra94: The only thing I have to recall is that why I only rely on `assigned` parameter. It was long time ago and I would need to deep dive in it again.

LMnet: That's a very valuable peace of information. Could you please add it as a comment somewhere? And is it possible to prove this behavior with the test?

MrKustra94: Okay, I took additional look.
Łukasz is right about assigned partitions - assigned callback is invoked only with new assigned partitons (Coming from https://github.com/apache/kafka/blob/38103ffaa962ef5092baffb884c84f8de3568501/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L417):
```java
        SortedSet<TopicPartition> ownedPartitions = new TreeSet<>(COMPARATOR);
        ownedPartitions.addAll(subscriptions.assignedPartitions());

        // should at least encode the short version
        if (assignmentBuffer.remaining() < 2)
            throw new IllegalStateException("There are insufficient bytes available to read assignment from the sync-group response (" +
                "actual byte size " + assignmentBuffer.remaining() + ") , this is not expected; " +
                "it is possible that the leader's assign function is buggy and did not return any assignment for this member, " +
                "or because static member is configured and the protocol is buggy hence did not get the assignment for this member");
        
        //Some code...
        Assignment assignment = ConsumerProtocol.deserializeAssignment(assignmentBuffer);

        SortedSet<TopicPartition> assignedPartitions = new TreeSet<>(COMPARATOR);
        assignedPartitions.addAll(assignment.partitions());

        //Some code...
        final AtomicReference<Exception> firstException = new AtomicReference<>(null);
        SortedSet<TopicPartition> addedPartitions = new TreeSet<>(COMPARATOR);
        addedPartitions.addAll(assignedPartitions);
        addedPartitions.removeAll(ownedPartitions);

        if (protocol == RebalanceProtocol.COOPERATIVE) {
            SortedSet<TopicPartition> revokedPartitions = new TreeSet<>(COMPARATOR);
            revokedPartitions.addAll(ownedPartitions);
            revokedPartitions.removeAll(assignedPartitions);

            log.info("Updating assignment with\n" +
                    "\tAssigned partitions:                       {}\n" +
                    "\tCurrent owned partitions:                  {}\n" +
                    "\tAdded partitions (assigned - owned):       {}\n" +
                    "\tRevoked partitions (owned - assigned):     {}\n",
                assignedPartitions,
                ownedPartitions,
                addedPartitions,
                revokedPartitions
            );
//..more code
```
but more magic happens when revoke callback is invoked (https://github.com/apache/kafka/blob/38103ffaa962ef5092baffb884c84f8de3568501/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L777):
```java
switch (protocol) {
                case EAGER:
                    // revoke all partitions
                    revokedPartitions.addAll(subscriptions.assignedPartitions());
                    exception = invokePartitionsRevoked(revokedPartitions);

                    subscriptions.assignFromSubscribed(Collections.emptySet());

                    break;

                case COOPERATIVE:
                    // only revoke those partitions that are not in the subscription any more.
                    Set<TopicPartition> ownedPartitions = new HashSet<>(subscriptions.assignedPartitions());
                    revokedPartitions.addAll(ownedPartitions.stream()
                        .filter(tp -> !subscriptions.subscription().contains(tp.topic()))
                        .collect(Collectors.toSet()));

                    if (!revokedPartitions.isEmpty()) {
                        exception = invokePartitionsRevoked(revokedPartitions);

                        ownedPartitions.removeAll(revokedPartitions);
                        subscriptions.assignFromSubscribed(ownedPartitions);
                    }

                    break;
            }
```
>And is it possible to prove this behavior with the test?
It is really really hard to reproduce it. It mainly happens when small amount of resources are assigned comparing to workload that must be consumed.

Issue #1099, opened 2022-11-09
TransactionalKafkaProducer don't commit offset with empty records


chernovsa:
According to java doc for `TransactionalProducerRecords`, 
`[records](records: Chunk[CommittableProducerRecords[F, K, V]])` can be empty, but nothing is said about committing offset. 
When we check documentation for `CommittableProducerRecords` we get:

`sealed abstract class CommittableProducerRecords[F[_], +K, +V] {`

  `// The records to produce. Can be empty to simply commit the offset. `
  `def records: Chunk[ProducerRecord[K, V]]`

   `// The offset to commit. `
`  def offset: CommittableOffset[F]`
`}`

That said: "The records to produce. Can be empty to simply commit the offset".
For example:
`val offsets = ???`
`val records = CommittableProducerRecords(Chunk.empty, offsets))`

So, It can confuse when you invoke `TransactionalProducerRecords::produce` with records of that type, where you have empty records, and non-empty offsets. Committing offset is expected, but nothing happens. More over, there is no indication about this case after method invokation, that offset have not committed. 
I believe that, my use case is common. 
Lets consider, that you have a Kafka Job application with usual flow like, consuming data from some topic, filtering data, transforming data, and producing to another topic.  U can see, that when filtering step is presented, then nothing to produce can occur in rare cases. And even this case, you still want to commit offset, to save your consuming progress.

chernovsa:
https://github.com/fd4s/fs2-kafka/pull/1100
Please consider to merge this pull request.

bplommer:
Fixed by #1100

PR #1100, opened 2022-11-09
1099 Fix produce method of TransactionalKafkaProducer, when method is invoked with empty records and non-empty offsets

chernovsa:
Fix produce method of TransactionalKafkaProducer, when method is invoked with empty records and not empty offsets. Offsets will be committed.
Closes issue https://github.com/fd4s/fs2-kafka/issues/1099


bplommer:
Thanks! This looks right to me. Could you run `scalafmtAll` and commit the result please?

chernovsa:
> Thanks! This looks right to me. Could you run `scalafmtAll` and commit the result please?

Yes, sure, will do.

bplommer:
Thanks!

PR #1111, opened 2022-11-24
Add partitionFor for given topic with KafkaProducer

### LINKED ISSUE #1054: Is there a way to check if the producer is ready to produce?
I have an app that exposes an http endpoint and produces into Kafka with `fs2-kafka`, however I would like to not make the app ready until I'm sure that the producer is ready to produce aka connected to the broker. 

Usually I could achieve that using something like `producer.partitionsFor(topic)` but that's not available within `fs2-kafka`s producers. Is there a way to achieve this?

sergio-margale:
Closes #1054

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -287,6 +287,9 @@ ThisBuild / mimaBinaryIssueFilters ++= {
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.ProducerSettings.withSerializers"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.vulcan.AvroSettings.*"),
       ProblemFilters.exclude[FinalMethodProblem]("fs2.kafka.vulcan.AvroSettings.*"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.kafka.KafkaProducerConnection.withSerializers"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.withSerializers"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.partitionsFor"),
 
       // private
         ProblemFilters.exclude[Problem]("fs2.kafka.vulcan.AvroSettings#AvroSettingsImpl.*")
--- a/build.sbt
+++ b/build.sbt
@@ -281,6 +281,7 @@ ThisBuild / mimaBinaryIssueFilters ++= {
 
       // package-private
       ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.kafka.KafkaProducer.from"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.kafka.KafkaProducer.from"),
 
       // sealed
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.ConsumerSettings.withDeserializers"),
--- a/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
@@ -6,16 +6,14 @@
 
 package fs2.kafka
 
+import cats.{Apply, Functor}
 import cats.effect._
 import cats.syntax.all._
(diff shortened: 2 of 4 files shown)


sergio-margale:
@bplommer could you have a look into this? Also, would this be ported if approved and merged into series 3.x?

sergio-margale:
@bplommer it should be done now 😄 
Please let me know if anything else is needed

modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
```diff
@@ -56,6 +54,10 @@ abstract class KafkaProducer[F[_], K, V] {
   def produce[P](
     records: ProducerRecords[P, K, V]
   ): F[F[ProducerResult[P, K, V]]]
+
+  def partitionsFor(
+    topic: String,
+  ): F[List[PartitionInfo]]
```
bplommer, line 60: This needs to be removed here so it's only defined in `KakfaProducer.PartitionFor` - I'm going to take the liberty of doing so.
```suggestion
```

bplommer:
Thanks! There's one remaining issue - there are some binary compatibility warnings in the build error (copied below). These are safe to ignore because they relate to a sealed trait - the exclusions (on the lines with "filter with") need to be added to the exclusions list in build.sbt (in the "sealed" section - https://github.com/fd4s/fs2-kafka/blob/5c0b4592f5b1d8488996aaaebd1639df7bc5a9a7/build.sbt#L285)

```
[error] fs2-kafka: Failed binary compatibility check against com.github.fd4s:fs2-kafka_2.13:2.4.0 (e:info.versionScheme=early-semver)! Found 3 potential problems (filtered 34)
[error]  * abstract method withSerializers(fs2.kafka.Serializer,fs2.kafka.Serializer)fs2.kafka.KafkaProducer#Metrics in class fs2.kafka.KafkaProducerConnection has a different result type in current version, where it is fs2.kafka.KafkaProducer#PartitionsFor rather than fs2.kafka.KafkaProducer#Metrics
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.kafka.KafkaProducerConnection.withSerializers")
[error]  * abstract method withSerializers(fs2.kafka.Serializer,fs2.kafka.Serializer)fs2.kafka.KafkaProducer#PartitionsFor in class fs2.kafka.KafkaProducerConnection is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.withSerializers")
[error]  * abstract method partitionsFor(java.lang.String)java.lang.Object in class fs2.kafka.KafkaProducerConnection is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.partitionsFor")
```

build.sbt
```diff
@@ -287,6 +287,9 @@ ThisBuild / mimaBinaryIssueFilters ++= {
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.ProducerSettings.withSerializers"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.vulcan.AvroSettings.*"),
       ProblemFilters.exclude[FinalMethodProblem]("fs2.kafka.vulcan.AvroSettings.*"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.kafka.KafkaProducerConnection.withSerializers"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.withSerializers"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.partitionsFor"),
```
bplommer, line 292: One more that came up in the scala 3 tests:
```suggestion
      ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.kafka.KafkaProducerConnection.partitionsFor"),
ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.kafka.KafkaProducer.from")
```

bplommer (approved):
Thanks for seeing this through!

sergio-margale:
Thank you for the hints!

PR #1150, opened 2023-02-17
Make instantiation of CommitTimeoutException lazy.

optician:
Hi! 
I lost about 1 day trying to figure out why there are a lot of CommitTimeoutException in jfr files (with stacktrace) and zero entries in application logs. So I suggest making it lazy to save time for others.


optician:
I don't like stacktrace performance in general, but in this case it does not seem to be a problem. I have a service with very limited resources, and the impact of stacktraces is negligible. With the default batch settings. I think this would be a problem with tiny batches, but kafka hits the performance hard for them anyway. So it's all about UX imho.

I doubt `raiseError` would be changed because it's a breaking change. Also cats-effect has [similar code](https://github.com/typelevel/cats-effect/search?q=%22F.defer%28F.raiseError%22). I think this is intentional behaviour. Perhaps this has been discussed already.

aartigao (approved):
Given that even CE authors use this pattern and it will be better for performance (even as a microoptimization), I think this deserves to be merged ✅ 

Thank you for your contribution! 🎉 🙇🏽

PR #1176, opened 2023-03-27
Revert "Change to publish on series branches"

bplommer:
This reverts commit 78fc4258c8df82919d8ae38f6221dc4b6318915c.

@vlovgr what was the purpose of this change? I'm reverting it now to publish v2.6 but we can change it again if needed.


vlovgr:
Yeah, I guess so. Also not sure how well it will work when having both 2.x and 3.x branches. The current setup was never designed with this in mind.

PR #1187, opened 2023-04-11
Preparatory changes for new CE cancelation semantics

### LINKED ISSUE #1181: CE 3.5.0 async cancelation semantics
I think fs2-kafka will require some changes to accommodate the [new async cancelation semantics](https://github.com/typelevel/cats-effect/releases/tag/v3.5.0-RC1) in the upcoming CE 3.5.0.

For instance [this timeout](https://github.com/fd4s/fs2-kafka/blob/v3.0.0-RC1/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala#L117) will probably be rendered useless. Also I suspect that currently this timeout only cancels the waiting on the callback but will leak the actual hanging commit, which is what those new semantics are guarding against.

bplommer:
Resolves #1181 by returning `Some(F.unit)` instead of `None` for the canceler of the uncancelable async commit operation.

Also uses configured timeout for sync commit, and renames some syntax to avoid clashing with a new method on `IO` in forthcoming CE3.5

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,7 @@ val scala213 = "2.13.10"
 
 val scala3 = "3.2.2"
 
-ThisBuild / tlBaseVersion := "2.5"
+ThisBuild / tlBaseVersion := "2.6"
 
 ThisBuild / tlVersionIntroduced := Map("3" -> "2.1.0")
 
--- a/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala
@@ -218,7 +218,17 @@ object KafkaProducer {
                 else promise.failure(exception)
               }
             )
-          }.as(F.fromFuture(F.delay(promise.future)))
+          }.as {
+            F.delay(promise.future).flatMap { fut =>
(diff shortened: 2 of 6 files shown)


Jasper-M:
The use of `fromFuture` in [KafkaProducer](https://github.com/fd4s/fs2-kafka/blob/v3.0.0/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L206) also renders `produce` uncancelable if I'm not mistaken. That may also be a case that cannot really be canceled in the underlying library. So then I guess you can either let it be uncancelable or allow leaky cancelation (`io.cancelable(IO.unit)` or somesuch?).

LMnet (approved):

vlovgr (dismissed):

bplommer:
Good point @Jasper-M - we should definitely allow cancellation of the inner effect that waits for the producer result, even if the cancelation is leaky. I've opened https://github.com/typelevel/cats-effect/pull/3597 to backport the new `fromFutureCancelable` to CE 3.4, but I think I'll inline it here in the meantime to get this out of the door.

Jasper-M:
I guess this is fixed, but github didn't close it for some reason?

PR #1226, opened 2023-07-15
wip: add sasl config poperties

wonk1132:
wonk in progress for adding config for SASL

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
@@ -10,6 +10,7 @@ import cats.Show
 import cats.effect.Resource
 import fs2.kafka.security.KafkaCredentialStore
 import org.apache.kafka.clients.consumer.ConsumerConfig
+import org.apache.kafka.common.config.SaslConfig
 import org.apache.kafka.common.requests.OffsetFetchResponse
 
 import scala.concurrent.ExecutionContext
@@ -417,6 +418,15 @@ object ConsumerSettings {
     override val recordMetadata: ConsumerRecord[K, V] => String,
     override val maxPrefetchBatches: Int
   ) extends ConsumerSettings[F, K, V] {
+
+    def withPlainSasl(usernameToken: String, passwordToken: String): ConsumerSettings[F, K, V] =
+      withProperties(
+        SaslConfigs.SASL_MECHANISM -> "PLAIN",
+        SaslConfigs.SASL_JAAS_CONFIG ->
+          s"""org.apache.kafka.common.security.plain.PlainLoginModule required username="${usernameToken}" password="${passwordToken}";""",
+        CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, "SASL_SSL"
+      )
+
     override def withCustomBlockingContext(ec: ExecutionContext): ConsumerSettings[F, K, V] =
       copy(customBlockingContext = Some(ec))
 


modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
```diff
@@ -417,6 +425,15 @@ object ConsumerSettings {
     override val recordMetadata: ConsumerRecord[K, V] => String,
     override val maxPrefetchBatches: Int
   ) extends ConsumerSettings[F, K, V] {
+
+    def withPlainSasl(usernameToken: String, passwordToken: String): ConsumerSettings[F, K, V] =
```
aartigao, line 429: I've never worked with SASL but from what I can see here in `PlainLoginModule` there's no reference to any token.

Why not removing that prefix? 👇🏽 

```suggestion
    def withPlainSasl(username: String, password: String): ConsumerSettings[F, K, V] =
```

modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
```diff
@@ -401,6 +403,12 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
     * Includes the credentials properties from the provided [[KafkaCredentialStore]]
     */
   def withCredentials(credentialsStore: KafkaCredentialStore): ConsumerSettings[F, K, V]
+
+  /**
+   * provides SASL credentials to Kafka host. A typical use case for this would be
+   * interacting with Confluent Cloud.
```
aartigao, line 409: I feel the second sentence is not needed. A user should be aware of what SASL is at this point (either by reading the docs or because he/she knows what's doing.

```suggestion
   * Provides SASL credentials to Kafka host.
```

docs/src/main/mdoc/consumers.md
```diff
@@ -122,6 +122,23 @@ ConsumerSettings(
  .withGroupId("group")
 ```
 
+### Sasl setting
```
aartigao, line 125:
```suggestion
### SASL settings
```

docs/src/main/mdoc/consumers.md
```diff
@@ -122,6 +122,23 @@ ConsumerSettings(
...
+```scala mdoc:silent
+ConsumerSettings(
+   keyDeserializer = Deserializer[IO, String],
+   valueDeserializer = Deserializer[IO, String]
+).withAutoOffsetReset(AutoOffsetReset.Earliest)
+ .withBootstrapServers("pkc-instanceName.us-west1.gcp.confluent.cloud:9092")
+ .withGroupId("group")
+ .withPlainSasl(
+   <idToken>,
+   <passwordToken>
```
aartigao, line 138: Same here, I don't think mentioning token is needed.

Also, here you are using `id` instead of `username`. I'd suggest to use the same convention everywhere.

aartigao (changes requested):
Thank you for your contribution! 🙇🏽 

Just some minor adjustments and the PR is ready to 🚀

modules/core/src/main/scala/fs2/kafka/ConsumerSettings.scala
```diff
@@ -401,6 +403,12 @@ sealed abstract class ConsumerSettings[F[_], K, V] {
...
+   */
+  def withPlainSasl(usernameToken: String, passwordToken: String): ConsumerSettings[F, K, V]
```
cwholmes, line 411: Could this be a new `KafkaCredentialStore` rather than being directly on the consumer settings? That would also add support for the producer and admin client.

PR #1227, opened 2023-07-18
Restore retries to default, since idempotency is guaranteed

soujiro32167:
The goal is to guarantee ordering, while preserving defaults as much as possible.

From fs2 docs:
> max.retries is set to 0, to avoid the risk of records being produced out-of-order. If we don't need to produce records in-order, then this can be set to some positive integer value. An alternative is to enable retries and use withMaxInFlightRequestsPerConnection(1) or withEnableIdempotence(true). The blog post [Does Kafka really guarantee the order of messages?](https://blog.softwaremill.com/does-kafka-really-guarantee-the-order-of-messages-3ca849fd19d2) provides more detail on this topic.

Consider the following [defaults](https://kafka.apache.org/documentation/#producerconfigs_enable.idempotence):
1. `retries: Int.MaxValue`
2. `enable.idempotence: true (if retries > 0 and max.in.flight.requests.per.connection <= 5)`
3. `max.in.flight.requests.per.connection: 5`

Since idempotance is enabled by default, the scenario where records arrive out of order is prevented.
Therefore, we can eliminate setting retries to 0

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
+++ b/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
@@ -450,30 +450,27 @@ private[kafka] object KafkaConsumerActor {
       */
     def resetSpilloverAfterPoll(
       spillover: Map[TopicPartition, Chunk[CommittableConsumerRecord[F, K, V]]]
-    ): State[F, K, V] =
-      if (spillover.isEmpty)
-        this
-      else {
-        require(spillover.forall(kv => partitionState.contains(kv._1)))
+    ): State[F, K, V] = {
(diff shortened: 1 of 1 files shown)


aartigao:
I'm concerned that this could break some existing codebases.

Imagine an scenario where users are aware of fs2-kafka not retrying on errors, and it's fine for them. Most probably the error handler and potential retry logic is set in user land. If we merge this change, we might broke that code.

TBH, I'm not sure how to proceed with this, because even if nothing changes at the API level, it does at runtime. I'm not sure if it's fine to accept this change in a minor version and just mention it in the release notes, or if this deserves a major release.

I'll let the more experienced people to decide on that @mpilquist

PR #1231, opened 2023-08-11
Remove VerifiableProperties issue workaround

peter-hazell:
The inclusion of this class causes problems when trying to use [embedded-kafka](https://github.com/embeddedkafka/embedded-kafka) (3.5.1), as it conflicts with the `VerifiableProperties` class in the kafka library and is used when starting the kafka server, which results in the following error:

```scala
Cause: java.lang.NoSuchMethodError: 'void kafka.utils.VerifiableProperties.<init>(java.util.Properties)'
[info]   at kafka.utils.VerifiableProperties$.apply(VerifiableProperties.scala:30)
[info]   at kafka.server.KafkaServer.<init>(KafkaServer.scala:117)
```

According to the comment, this class was originally added as a workaround for https://github.com/lampepfl/dotty/issues/13523 and https://github.com/confluentinc/schema-registry/issues/553. Since then, both of these issues appear to be closed/resolved and therefore it would seem that this class can be removed, which will allow us to use embedded-kafka.


LMnet:
@bcarter97 looks like something goes wrong on CI for scala 3: https://github.com/fd4s/fs2-kafka/actions/runs/6257945535/job/16991091067

bcarter97:
AFAIK this file is a dummy to find a `kafka.utils.VerifiableProperties` on the classpath. Now it's been fixed upstream, this isn't required. Binary compat will fail as we're removing a file without a replacement, but it's private and nothing is dependent on it in-code so I don't think it's possible for anything to actually break?

aartigao:
@LMnet Is fixed in #1245 

I put it there because I was touching SBT stuff, but can create another PR explicitly for this (although I don't think it's necessary 🙃).

Issue #1230, opened 2023-08-10
Stuck partitions - question about the pause/resume logic


simonpetty:
Hey.

We have a topic with 50 partitions, with consumers scaled out to a relatively small number of hosts (3). Sporadically, we get a situation where one particular partition gets stuck and does not commit any new offsets. 

We see this in the logs very rapidly after it's become stuck:

`Skipping fetching records for assigned partition <mypartition> because it is paused`

The last couple of FS2 Kafka logs mentioning the stuck partition, before it was stuck, say this:

`Completed fetches with records for partitions [ <mypartition> -> { first: 45677489, last: 45677495 }, .... ]`

Followed by logs that never mention the partition again:

`Current state [State(fetches = Map(... a number of other partitions, but not <mypartition> ... ), ...)]`

In trying to figure out what's going on, I ended up looking at this bit of the code:

```
def pollConsumer(state: State[F, K, V]): F[ConsumerRecords] =
      withConsumer
        .blocking { consumer =>
          val assigned = consumer.assignment.toSet
          val requested = state.fetches.keySetStrict
          val available = state.records.keySetStrict

          val resume = (requested intersect assigned) diff available
          val pause = assigned diff resume

          if (pause.nonEmpty)
            consumer.pause(pause.asJava)

          if (resume.nonEmpty)
            consumer.resume(resume.asJava)

          consumer.poll(pollTimeout)
        }
        .flatMap(records)
 ```

Filling in all the possible combinations, I think we get these possible outcomes:

<img width="556" alt="Screenshot 2023-08-10 at 09 01 54" src="https://github.com/fd4s/fs2-kafka/assets/223079/59f03030-930b-4fde-867f-3f369f3e97ff">

I've highlighted the state that I think we're in, and given that the java kafka library is saying it will skip fetching for paused partitions, so I'm a little confused how the application is expected to resume once it's in this state.

If I pull down the FS2 Kafka codebase, and replace the `intersect` with `union` the KafkaConsumerSpec tests still pass, and it would cause our highlighted scenario to be in resume, but I don't know what the consequences of that would be!

It feels like this should be a more widespread issue given how central this code is, and how long it's been like this for, so I bet i'm missing something.

Thanks

jarrodcodes:
Just curious because we are having a similar issue, which version of fs2-kafka are you on? We are on 2.6.1, wanted to know if this was also happening with 3.x

tpalmer99:
[...]

PR #1238, opened 2023-09-06
Retire FakeFiber

biochimia:
FakeFiber was used by KafkaConsumer to manage:

1. the consumerActor fiber, processing requests (including polls);
2. the pollScheduler fiber, scheduling poll requests (subject to backpressure);
3. a fiber combining the above two fibers.

Compared to a regular fiber, FakeFiber offered a method to combine two fibers by racing them one against the other. The semantics were similar to the race method for effects, but operating at the fiber level.

In KafkaConsumer, FakeFiber was used with cancellation effects that returned immediately (i.e., fiber.cancel.start.void). In addition the fiber outcome was relayed to KafkaConsumer.awaitTermination.

With this change FakeFiber is replaced with an Async[F].race of the underlying effects. This is managed in the new method startBackgroundConsumer, which builds upon effects assembled by runConsumerActor and runPollScheduler. These effects are unwrapped from the previous fibers.

As before, cancellation of the consumer effect is only waited on in awaitTermination, where any errors are propagated.

Compared to the original behaviour of FakeFiber.combine, starting with cats-effect 3.5.0, cancellation of one of the consumer effects will lead to cancellation of both, as per changes introduced with https://github.com/typelevel/cats-effect/pull/3453. I'm not sure how cancelation would come into play here, but the behaviour change looks appropriate: without one of the racing fibers KafkaConsumer would not be functional (no polls scheduled, or no requests/polls processed).


biochimia:
As motivation for doing this, I'm running down different code paths trying to understand how partitions may get "lost", in relation to #1230. To be clear, I have _not_ identified any issue with `FakeFiber` that would relate to that.

I do note that FakeFiber was introduced to support the upgrade to cats-effect 3.x, and that its behaviour seems to be well supported by racing the effects directly, instead. With this change the goal is to remove a little bit of legacy and hopefully reduce cognitive load as well.

aartigao (approved):
LGTM 👍🏽

PR #1246, opened 2023-09-24
Bump CE and fs2 to the latest version

### LINKED ISSUE #1234: upgrade 3.x to CE 3.5.1 / fs2 3.8.0
Hello! I'm working on some dependency updates for $work and fs2-kafka is keeping us on cats-effect 3.4.11 (aka before the async cancellation changes) and fs2 3.6.1

I noticed that CE + fs2 upgrades are ignored in the scala-steward dependencies for 3.x, and I found a TODO regarding the update:
https://github.com/fd4s/fs2-kafka/blob/e4376d1df82c786c11e968b7945639a7addadbdc/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L208-L209

Is there a plan or anything in progress for that version bump yet? If not (and if it would be helpful) I could see how far I get and open up a PR? Since they were explicitly ignored I didn't want to jump in, in case I'm missing important context

aartigao:
Fixes #1234.

**NOTE:** I had to bump Scala 3 lib because [since 3.8.0](https://github.com/typelevel/fs2/releases/tag/v3.8.0) fs2 uses 3.3.x

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,12 +28,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.17, 2.13.10, 3.2.2]
+        scala: [2.12.17, 2.13.10, 3.3.1]
         java: [temurin@8, temurin@17]
         exclude:
           - scala: 2.12.17
             java: temurin@17
-          - scala: 3.2.2
+          - scala: 3.3.1
             java: temurin@17
     runs-on: ${{ matrix.os }}
     steps:
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,18 +1,6 @@
 pullRequests.frequency = "14 days"
 
 updates.ignore = [{
-  groupId = "org.typelevel",
-  artifactId="cats-effect"
-},{
-  groupId = "org.typelevel",
-  artifactId="cats-effect-laws"
-},{
-  groupId = "org.typelevel",
-  artifactId="cats-effect-testkit"
-},{
-  groupId = "co.fs2",
-  artifactId="fs2-core"
-},{
   groupId = "com.dimafeng"
 },{
   groupId = "org.scalameta",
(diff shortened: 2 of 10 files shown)


aartigao:
In addition to the changes suggested, while removing the `TODO` in `produceRecord[F[_], K, V]` method, I noticed the usage of `Promise` there inside `Callback`.

Consider the following code that uses directly the `Producer` API method with signature `Future<RecordMetadata> send(ProducerRecord<K, V> record);` and waits asynchronously to the result of the returned `Future` using `F.async_` 👇🏽 

```scala
private[kafka] def produceRecord[F[_], K, V](
  keySerializer: KeySerializer[F, K],
  valueSerializer: ValueSerializer[F, V],
  producer: KafkaByteProducer,
  blocking: Blocking[F]
)(
  implicit F: Async[F]
): ProducerRecord[K, V] => F[F[(ProducerRecord[K, V], RecordMetadata)]] =
  record =>
      blocking {
        producer.send(javaRecord)
      }.map { future =>
        F.async_ { cb =>
          Try(future.get) match {
            case Success(metadata)               => cb(Right((record, metadata)))
            case Failure(ex: ExecutionException) => cb(Left(ex.getCause))
            case Failure(exception)              => cb(Left(exception))
          }
        }
      }
    }
```

According to `async_` documentation:

```
This function can be thought of as a safer, lexically-constrained version of `Promise`,
where `IO` is like a safer, lazy version of `Future`.
```

What do you think about it? TBH I'm not really sure if there's any benefit on using `async_` or if it's actually worse 🤷🏽 (the `Try(future.get)` part is the one that concerns me).

cc @vlovgr

vlovgr:
> What do you think about it? TBH I'm not really sure if there's any benefit on using `async_` or if it's actually worse 🤷🏽 (the `Try(future.get)` part is the one that concerns me).

@aartigao With the new cancellation semantics in cats-effect 3.5.0, `F.async_` is uncancelable. `Try(future.get)` is also blocking, so I'd suggest we keep what you already have in this pull request. 👍

vlovgr (approved):

PR #1281, opened 2023-11-30
Add chunk-based consumer API

L7R7:
This adds an API to make the pattern of chunk-based consumers a first-class concept. The idea keeps popping up in the typelevel discord, and we've been successfully using it at $WORK for a while now, so it makes sense to add it.

# General idea

The pattern aims at helping users to implement consumers without auto-commit to write their code without having to do too much work in order to achieve correct offset committing (no offsets must be lost, offsets should be committed only after messages have been processed, etc).
It achieves that by switching from processing messages in a `Stream[F, CommittableConsumerRecord[F, K, V]]` to processing them chunks: `Chunk[ConsumerRecord[K, V]] => F[Unit]`. After each chunk, the offsets of the chunk are committed.

This has a couple of advantages (summarizing [what Fabio said on the Discord](https://discord.com/channels/632277896739946517/632310980449402880/1179069659128008775)) :
* Consumers can write their business logic without having to care about offsets
* batching several messages is very straightforward, which is often practical (e.g. batch writes to the db)
* concurrency and filtering can be done liberally without having to worry about incorrect commits

# open questions

* Do we want to mirror all methods from the `KafkaConsume` trait in the new trait? I only added one of them so far to show the concept, the rest could be implemented in the same fashion.
* Instead of `Unit`, the processor for the chunks could use a return type of `F[CommitNow]`, where `CommitNow` is an intention revealing equivalent of `Unit` that also makes it more clear that afterwards no processing should be done on the records. Do we want that?
* Instead of only processing single chunks, the library could also offer (additional?) ways that take care of things like:
  * deduplicating messages in a chunk by key (I feel this may be too much for the library and might be confusing for some consumers, and it's also relatively easy to implement consumer side)
  * allowing to process messages concurrently, using concurrency per key (this is a bit more tricky to implement, so it might be a good idea to provide it in an opt-in way for those consumers who want it but don't want to implement it themselves). If we decide to allow both the "simple" chunk-based approach as well as the "concurrent-by-key" approach, I'd be happy to add that after this PR is done.

# todos

There's still a lot of work to do, but I wanted to get some early feedback on the general concept before doing the busy work.

- testing
- docs
- add the missing methods to make the API complete and consistent to the rest

I'm looking forward to receiving some feedbacks and thoughts on this idea!

### DIFF
--- a/docs/src/main/mdoc/consumers.md
+++ b/docs/src/main/mdoc/consumers.md
@@ -12,6 +12,7 @@ import scala.concurrent.duration._
 
 import cats.effect._
 import cats.syntax.all._
+import fs2._
 import fs2.kafka._
 import fs2.kafka.consumer.KafkaConsumeChunk.CommitNow
 ```
--- a/docs/src/main/mdoc/consumers.md
+++ b/docs/src/main/mdoc/consumers.md
@@ -272,6 +272,45 @@ object ConsumerMapAsyncExample extends IOApp.Simple {
 
 Offsets commits are managed manually, which is important for ensuring at-least-once delivery. This means that, by [default](#default-settings), automatic offset commits are disabled. If you're sure you don't need at-least-once delivery, you can re-enable automatic offset commits using `withEnableAutoCommit` on [`ConsumerSettings`][consumersettings], and then ignore the [`CommittableOffset`][committableoffset] part of [`CommittableConsumerRecord`][committableconsumerrecord], keeping only the [`ConsumerRecord`][consumerrecord].
 
+### Working on `Chunk`
+
+Most use cases that require at-least-once delivery make it necessary to commit the offset of messages only after the message has been successfully processed. Implementing this correctly can be challenging, especially when the business logic requires advanced data manipulation with concurrency, batching, filtering and the like:
+
+- When consuming multiple messages from the same partition concurrently, a consumer might lose messages if the commits happen out of order and one of the older messages can't be processed and has to be retried.
+- When filtering messages, it's important to still commit the offset of the filtered message because if this message is the latest one on its partition, it will get re-sent infinitely.
+- For performance reasons, it makes sense to batch the offsets when committing them.
+
+The recommended pattern for these use cases is by working on the `Chunk`s of records that are part of the `Stream`. The library supports that with the `consumeChunk` method:
+
+```scala mdoc:silent
+object ConsumerChunkExample extends IOApp.Simple {
+  val run: IO[Unit] = {
+    def processRecords(records: Chunk[ConsumerRecord[String, String]]): IO[CommitNow] =
+      records.traverse(record => IO.println(s"Processing record: $record")).as(CommitNow)
+
+    KafkaConsumer.stream(consumerSettings)
+      .subscribeTo("topic")
+      .consumeChunk(processRecords)
+  }
+}
+```
+
+Note that this method uses `partitionedStream`, which means that all the partitions assigned to the consumer will be processed concurrently.
+
+As a user, you don't have to care about the offset commits, all you have to do is implement a function that processes all records in the `Chunk`, and return a `IO[CommitNow]`. After this action finished, the offsets for all messages in the `Chunk` will be committed. `CommitNow` is basically the same as `Unit`, but helps in making it clear when the processing of messages has been finished.
+
+This brings several benefits for the most use cases:
+
+- **Correctness:** You can focus on implementing your business logic, without having to worry about offset commits or propagating the correct offsets through your code. Offsets are committed correctly afterwards.
+- **Performance:** Typical performance improvements are bulk-writes to a database, or using concurrency to speed things up. These patterns can be used liberally when working on the records in a `Chunk`, without having to sacrifice correctness.
+- **Flexibility:** Besides using batching and concurrency, you might want to filter out messages, or process them in a different order than they appear on the partitions. As long as you work on a single `Chunk` and make sure that the processing is finished when you return `CommitNow`, you can do all that.
+
+If the chunk size doesn't fit your needs, the first way to start tuning is the `max.poll.records` config property of your consumer.
+
+### Committing manually
+
+If `consumeChunk` doesn't work for you, you can always commit your offsets manually.
+
 Offset commits are usually done in batches for performance reasons. We normally don't need to commit every offset, but only the last processed offset. There is a trade-off in how much reprocessing we have to do when we restart versus the performance implication of committing more frequently. Depending on our situation, we'll then choose an appropriate frequency for offset commits.
 
 We should keep the [`CommittableOffset`][committableoffset] in our `Stream` once we've finished processing the record. For at-least-once delivery, it's essential that offset commits preserve topic-partition ordering, so we have to make sure we keep offsets in the same order as we receive them. There is one convenience function for the most common batch committing scenario, `commitBatchWithin`.
--- a/docs/src/main/mdoc/consumers.md
+++ b/docs/src/main/mdoc/consumers.md
@@ -13,6 +13,7 @@ import scala.concurrent.duration._
 import cats.effect._
 import cats.syntax.all._
 import fs2.kafka._
+import fs2.kafka.consumer.KafkaConsumeChunk.CommitNow
 ```
 
 ## Deserializers
--- a/docs/src/main/mdoc/consumers.md
+++ b/docs/src/main/mdoc/consumers.md
@@ -275,9 +275,9 @@ Offsets commits are managed manually, which is important for ensuring at-least-o
 
 ### Working on `Chunk`
 
-Most use cases that require at-least-once delivery make it necessary to commit the offset of messages only after the message has been successfully processed. Implementing this correctly can be challenging, especially when the business logic requires advanced data manipulation with concurrency, batching, filtering and the like:
+Use cases that require at-least-once delivery make it necessary to commit the offset of messages only after the message has been successfully processed. Implementing this correctly can be challenging, especially when the business logic requires advanced data manipulation with concurrency, batching, filtering and the like:
 
-- When consuming multiple messages from the same partition concurrently, a consumer might lose messages if the commits happen out of order and one of the older messages can't be processed and has to be retried.
+- When consuming multiple messages from the same partition concurrently, a consumer might lose messages if the commits happen out of order and a message that is not the last one on its partition can't be processed and has to be retried.
 - When filtering messages, it's important to still commit the offset of the filtered message because if this message is the latest one on its partition, it will get re-sent infinitely.
 - For performance reasons, it makes sense to batch the offsets when committing them.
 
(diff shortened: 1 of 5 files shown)


L7R7:
After adding the missing methods smiliar to the ones from the `KafkaConsume` trait, I'm wondering if it makes sense to simply "mirror" the signatures of the underlying methods.  `streamChunk` is obvious, but with `partitionedStreamChunk` and maybe even `partitionsMapStreamChunk`, I'm not sure if we should expose the different streams to the user because I don't know what a good scenario would look like in which the user may want to do something with the streams after each chunk is processed and commited.
At the moment, I think it's a tradeoff between symmetry to `KafkaConsume` and more convenience for the user.

aartigao:
Nice work! Let's start the discussion 🧵 

This addition _looks good_, because it leverages the logic to the library, but it also means we're being more strict on the patterns we support. For example, our $WORK use case doesn't involve committing offsets, but we're working with `Chunk` for performance reasons. Does that mean that the lib should support that too? Or this is a user-land decision? 🤷🏽 

So far, in my experience, working with `Stream` directly gives you the building blocks for crafting more complex patterns, like the one you expose.

The discussion this PR brings, philosophically, is somewhat like: do we think there's a canonical pattern for Kafka consumption around the `Chunk[Record] => F[Unit]` pattern?

And to be honest I don't know. I've mixed feelings, on one hand, by looking at the code, it's so simple that adding it to the library doesn't disrupt the rest of the API, but on the other hand, if it's that simple, why not having in in user-land? Otherwise looks to me that we maybe want this library to become more like a framework? Because we are loosing (not actually loosing, because the current API remains untouched, but you get the idea) the `Stream` foundations in favor of a more constrained API.

What I agree on is to properly document this use case, so the question doesn't pop up again in Discord.

I'd love to see what the original designers think about it 👀

SystemFw:
My 2c:
I think it's worth adding, because it's close to universally useful when working with Kafka for OLTP rather than OLAP use case. In this case, what we're adding is not the small amount of simple code, but the _expertise_ that this how you can achieve performance, flexibility and correctness.
In my extensive experience working with teams using fs2-kafka, that expertise is lacking and at best you get correct code which leaves a lot of performance on the table, and at worst you get incorrect code. The questions in Discord show a similar pattern.

However, what I would not do is add the whole set of variations mirroring the full api, I'd really just add the one method, otherwise we're back to the issue of people having to navigate which one to use. If you know enough that you need a variation, you can use the low-level Stream api directly.

Asides:
 - in the one method we do add, the return type should be `F[Nothing]`, rather than `Stream`.
 - you can avoid traversing the `Chunk` twice with `map`, by foldMapping it  instead to a tuple of offsets and records

L7R7:
Thank you for your comments, really appreciate it.

I think it's a good idea to only add one, biased, method to give the users an obvious way of using the patterns that works well for most cases. I'll proceed in that direction and will also start to add some docs for it. These docs will also mention that this is the first place where the library automatically commits offsets, but not using auto-commit (@aartigao thank you for mentioning the auto-commit use cases, I didn't think about that initially).

@SystemFw regarding your points to the implementation details:
* traversing only once over the Chunk totally makes sense, I couldn't find a way to use `foldMap` because there's no monoid instance for `CommittableOffsetBatch`, but it works fine with a normal `foldLeft`
* I don't quite understand yet how I turn a `Stream[Unit]` into a `Stream[Nothing]`, or a `F[Unit]` into a `F[Nothing]`. I think you chose `Nothing` to indicate that this action doesn't emit anything, and on `Stream` that is a good way to indicate that the stream won't emit any elements (since a `Stream[Unit]` might emit several elements), but will just finish at some point. I'm not sure if that applies to `F[Unit]` vs `F[Nothing]` as well. I came up with `.flatMap(_ => Stream.exec(F.unit)).compile.onlyOrError`, but I'm not sure if that's what you intended. Am I missing something?

SystemFw:
It should be an `F[Nothing]` (so, not a `Stream`, nor an `IO` of something else), because it represents a process that doesn't terminate, doesn't emit any results, and doesn't need to be composed as a `Stream`.
Just `compile.drain >> F.never`

L7R7:
Oh, I see. But that relies on the assumption that the previous stream never finishes, right? If the stream would terminate without error, the whole action wouldn't terminate because of the `never`

L7R7:
Should we use `consumeChunk` in the [quick example in the docs](https://fd4s.github.io/fs2-kafka/docs/quick-example) to nudge people in the "right direction" right from the beginning?

L7R7:
I'd say this is ready to be reviewed, the only questions I have is the one above regarding the `F.never`, and whether I should squash the commits into one single commit (I think it should be squashed, I left it unsquashed because it might be easier to review then)

modules/core/src/main/scala/fs2/kafka/consumer/KafkaConsumeChunk.scala
```diff
@@ -0,0 +1,88 @@
...
+      _.chunks.evalMap(consumeChunk(processor))
+    )
+    .parJoinUnbounded
+    .compile
+    .drain >> F.never
+
+  private def consumeChunk(processor: Chunk[ConsumerRecord[K, V]] => F[CommitNow])(
+    chunk: Chunk[CommittableConsumerRecord[F, K, V]]
+  )(implicit F: Monad[F]): F[Unit] = {
+    val (offsets, records) = chunk.foldLeft(
```
o0hgt0o, line 67: Great work! I'm not an expert in fs2, but perhaps this will be more memory efficient than allocating chunks to each element?
```suggestion
    val (offsets, records) = chunk
      .mapAccumulate(CommittableOffsetBatch.empty)((offsetBatch, cRecord) =>
        (offsetBatch.updated(cRecord.offset), cRecord.record)
      )
```

L7R7: Thank you for the suggestion, I wasn't aware of `mapAccumulate`. To be honest, I can't really tell if this would be more memory efficient, but from the signature I guess it should definitely be `mapAccumulate` and not `foldLeft`

docs/src/main/mdoc/quick-example.md
```diff
@@ -10,52 +10,36 @@ Following is an example showing how to:
...
         .withGroupId("group")
 
     val producerSettings =
-      ProducerSettings[IO, String, String].withBootstrapServers("localhost:9092")
+      ProducerSettings[IO, String, String]
+        .withBootstrapServers("localhost:9092")
+
+    def processRecords(producer: KafkaProducer[IO, String, String])(records: Chunk[ConsumerRecord[String, String]]): IO[CommitNow] = {
+      val producerRecords = records.map(consumerRecord => ProducerRecord("topic", consumerRecord.key, consumerRecord.value))
+      producer.produce(producerRecords).flatten.as(CommitNow)
```
lacarvalho91, line 31: does this mean that the size of the producer batches is always the size of the chunk? since the first effect is to add the records to the batch right? so we always wait for the chunk to be sent straight away after adding them to a batch to send?

L7R7: Yes. The outer effect completes after the messages are put in the buffer of the underlying producer, the inner effect completes when the messages are definitely published to Kafka.
From a performance perspective, it's okay to use `flatten` here because we're working on a chunk and not on individual messages. Waiting for the messages to have been published is important for correctness.

modules/core/src/main/scala/fs2/kafka/consumer/KafkaConsumeChunk.scala
L7R7: @o0hgt0o Thanks again for your suggestion, I changed it to `mapAccumulate`, I like that more than the `foldLeft`

modules/core/src/main/scala/fs2/kafka/consumer/KafkaConsumeChunk.scala
```diff
@@ -0,0 +1,96 @@
...
+  final def consumeChunk(
+    processor: Chunk[ConsumerRecord[K, V]] => F[CommitNow]
+  )(implicit F: Concurrent[F]): F[Nothing] = partitionedStream
+    .map(
+      _.chunks.evalMap(consume(processor))
+    )
+    .parJoinUnbounded
+    .compile
+    .drain
+    .flatMap[Nothing](_ => F.never)
```
L7R7, line 69: I'm not sure if this is the right way to solve the compile error. I think the ideal solution would be making the `_` type parameter of the `F` in `KafkaConsumeChunk`  covariant, but that would require it to be covariant in `KafkaConsumer` as well, which would probably rather invasive

aartigao: I was going to propose:

```scala
    .parJoinUnbounded
    .drain
    .compile
    .lastOrError
```

WDYT?

aartigao: But I can't test it due to compilation issues with the tests.

L7R7: I'm not sure if the types will line up with `lastOrError`, but I'll give it a try. I'm currently working on fixing the compilation issues in the tests (I only tried it locally with Scala 3 and didn't think about different Collection APIs in the 2.x versions)

L7R7: The tests should be fixed, compiles on my machine :tm: with 2.12.18, 2.13.10, and 3.3.0.

I tried `lastOrError`, but I think this leads to the same issue with the initial `>> F.never` variant

L7R7: Technically, `.drain >>[Nothing] F.never` would work as well, but with `flatMap` I think it's a bit less confusing syntax-wise

aartigao: >I tried lastOrError, but I think this leads to the same issue with the initial >> F.never variant

Can you share the code? I've tried it and works (compiles and tests pass) 🤔

docs/src/main/mdoc/quick-example.md
```diff
@@ -10,52 +10,36 @@ Following is an example showing how to:
...
-            .flatMap { producer =>
-              offsetsAndProducerRecords
-                .evalMap { case (offset, producerRecord) =>
-                  producer.produce(producerRecord).map(_.as(offset))
-                }
-                .parEvalMap(Int.MaxValue)(identity)
-            }
-        }
-        .through(commitBatchWithin(500, 15.seconds))
+      KafkaProducer.stream(producerSettings).flatMap { producer =>
```
aartigao, line 35:
```suggestion
      KafkaProducer.stream(producerSettings).evalMap { producer =>
```

docs/src/main/mdoc/consumers.md
```diff
@@ -13,6 +13,7 @@ import scala.concurrent.duration._
 import cats.effect._
 import cats.syntax.all._
 import fs2.kafka._
+import fs2.kafka.consumer.KafkaConsumeChunk.CommitNow
```
aartigao, line 16:
```suggestion
import fs2._
import fs2.kafka._
import fs2.kafka.consumer.KafkaConsumeChunk.CommitNow
```

docs/src/main/mdoc/quick-example.md
```diff
@@ -10,52 +10,36 @@ Following is an example showing how to:
 - use `commitBatchWithin` to commit consumed offsets in batches.
 
 ```scala mdoc
-import scala.concurrent.duration._
-
 import cats.effect.{IO, IOApp}
 import fs2.kafka._
+import fs2.kafka.consumer.KafkaConsumeChunk.CommitNow
```
aartigao, line 15:
```suggestion
import fs2._
import fs2.kafka._
import fs2.kafka.consumer.KafkaConsumeChunk.CommitNow
```

aartigao (commented):
Regarding docs, without these suggestions, `docs / run` is failing

modules/core/src/main/scala/fs2/kafka/consumer/KafkaConsumeChunk.scala
L7R7: You're right, sorry! I missed the `drain` before `compile`.

I like this more than what's currently implemented, because it addresses the point I brought up earlier: https://github.com/fd4s/fs2-kafka/pull/1281#issuecomment-1849729779 If for some reason the stream terminates, the whole action ends. With `F.never`, it "ends" silently.

We could bikeshed if it's `onlyOrError` or `lastOrError` now, but I think that doesn't matter

aartigao: > We could bikeshed if it's `onlyOrError` or `lastOrError` now, but I think that doesn't matter

Exactly! I'm also uncertain of what to use (`only` vs `last`), so whatever you choose will be good :)

L7R7: I went with `only`, because `last` has the notion of "drop some elements" and in that case `only` is closer to "there won't be any elements in the stream"

aartigao (approved):
I'm super grateful of this type of PR:

1. Small # of files
2. Single feature
3. Well documented

Bravo @L7R7 !

Enough time has passed since the creation and I'll be merging it soon.

Congrats to everyone involved!

L7R7:
@aartigao thank you for the review and thank you for helping me get through the last details yesterday.
I enjoyed working on that PR :smiley:

Issue #94, opened 2019-02-26
Expose additional consumer functionality


vlovgr:
The following functions could be added to `KafkaConsumer` to expose additional functionality provided by the Java Kafka consumer. Note that already implemented functions at the time of writing are not included. Also, functions for comitting offsets and pausing/resuming partitions are not included.

- [ ] `assign(Set[TopicPartition]): F[Unit]`
- [ ] `committed(TopicPartition): F[OffsetAndMetadata]`
- [ ] `committed(TopicPartition, FiniteDuration): F[OffsetAndMetadata]`
- [ ] `listTopics: F[Map[String, List[PartitionInfo]]]`
- [ ] `listTopics(FiniteDuration): F[Map[String, List[PartitionInfo]]]`
- [ ] `metrics: F[Map[MetricName, Metric]]`
- [ ] `offsetsForTimes(Map[TopicPartition, Long]): F[Map[TopicPartition, OffsetAndTimestamp]]`
- [ ] `offsetsForTimes(Map[TopicPartition, Long], FiniteDuration): F[Map[...]]`
- [ ] `partitionsFor(String): F[List[PartitionInfo]]`
- [ ] `partitionsFor(String, FiniteDuration): F[List[PartitionInfo]]`
- [ ] `paused: F[Set[TopicPartition]]`
- [ ] `subscribe(Regex, OnRebalance): F[Unit]`
- [ ] `subscribe[G[_]: Reducible](G[String], OnRebalance): F[Unit]`
- [ ] `subscription: F[Set[String]]`
- [ ] `unsubscribe: F[Unit]`

bplommer:
> Should `poll` & co. also be on this list? It is a very low-level functionality, but it seems to be required for certain usages, such as seeking to the earliest offset (`seekToBeginning`) without having to create a new `consumer.stream` to force the seeking to take effect?

This sounds reasonable to me. What do you think @vlovgr ?

bplommer:
Though on second thought, won't that take effect when polling next happens anyway assuming you already have a `consumer.stream` open?

valenterry:
> Though on second thought, won't that take effect when polling next happens anyway assuming you already have a `consumer.stream` open?

In my case I would close the stream for that

bplommer:
> > Though on second thought, won't that take effect when polling next happens anyway assuming you already have a `consumer.stream` open?
> 
> 
> 
> In my case I would close the stream for that

I don't think I'm following. Could you post a small code example?

[...]

PR #1285, opened 2023-12-16
Support consumer offsets for times and list topics

cwholmes:
This change adds functionality for the consumer for `offsetsForTimes` and `listTopics`. These are part of #94.


cwholmes:
@vlovgr @bplommer @aartigao I'm not sure who the correct people to tag for this would be, but I was wondering if this functionality could be reviewed/added. I have a project that is trying to use fs2-kafka and needs the offsets for times functionality. Please let me know if you are not the correct people.

chaddaksha:
@vlovgr @bplommer @aartigao Just bumping this thread to see if we can have this incorporated. Thanks in advance!

aartigao:
Hey, probably next week I can take a look! (quite busy these days with the newborn and all the celebrations...)

chaddaksha:
Thanks! No worries. Happy New Year :)

aartigao:
CI is not passing for Scala 2.12.

Please, add a compatibility converter for it in `fs2.kafka.internal.converters` object 🙏🏽

cwholmes:
I replaced the usage with something that works for all scala versions. The ci should hopefully be fixed now.

aartigao (approved):
LGTM!

cwholmes:
@aartigao Sorry. Had to fix an imports issue in a test, and updated a case to handle nulls more appropriately. Thanks for reviewing.

modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
```diff
@@ -613,6 +613,40 @@ object KafkaConsumer {
...
 
+      override def offsetsForTimes(
+        timestampsToSearch: Map[TopicPartition, Long]
+      ): F[Map[TopicPartition, OffsetAndTimestamp]] =
+        withConsumer.blocking {
+          _.offsetsForTimes(
+              timestampsToSearch.asJava.asInstanceOf[util.Map[TopicPartition, java.lang.Long]]
+            )
+            // Filter empty/missing partition null values for more idiomatic scala
+            .toMapNoNullValues
```
Jasper-M, line 624: Perhaps it should return a `Map[TopicPartition, Option[OffsetAndTimestamp]]` instead of filtering out the null entries?

cwholmes: I can change it to that if that is desired.

cwholmes: Updated to `Option` rather than removing the nulls

aartigao:
@cwholmes Can you run `sbt prePR` to fix the headers' year? 🙏🏽 

After that, if you don't have more changes, I'd be more than happy to merge it and build a release 🚀

aartigao (approved):

cwholmes:
It seems the files I added needed the dates on the headers updated. I have added that and things should be working again.

aartigao:
Thanks for the contribution! ;)

PR #1340, opened 2024-08-05
KafkaProducer.produce fails fast

rodrigo-molina:
## Description
Introduce `KafkaProducer.failFastProduce` producer configuration:

https://github.com/fd4s/fs2-kafka/blob/aed7ab3aeda58ac3b32438c5c22fb91c86ee2426/modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala#L218-L228

## Proposed solution
When `failFastProduce` is set to true, each `fs2.kafka.producer.produce` call of `ProducerRecords` creates a `scala.concurrent.Promise[Throwable]`. This Promise resolution is raced against calling `org.apache.kafka.clients.producer.KafkaProducer.send` per record. Any failing callback will complete the promise with the failure, canceling the rest of the pending `org.apache.kafka.clients.producer.KafkaProducer.send` and in-flight callbacks.

https://github.com/fd4s/fs2-kafka/blob/2559d96ce326f9f2ffc2936c1b9b5f3086a7829c/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L186-L198

## Explanation
`fs2.kafka.KafkaProducer.produce` returns F[F[...]], where the first effect invokes the underlying `org.apache.kafka.clients.producer.KafkaProducer.send` and attaches promises to the corresponding callbacks. The second effect is waiting for the callbacks’ resolution. In use cases where all the records in the batch need to be confirmed, any callback failure will cause the final effect resolution to fail, achieving the same outcome sooner.

In non-transient errors such as invalid username, password, or non-existent topics, `org.apache.kafka.clients.producer.KafkaProducer.send` is blocked for max.block.ms, default to 60 seconds. For example, a `fs2.kafka.producer` attempting to produce N `ProducerRecords` will take at least N * 60 seconds to fail. In other words, faulty Kafka producers will take considerable time to fail by repeatedly calling brokers for metadata synchronization.

See: 
- [max.block.ms default ](https://github.com/apache/kafka/blob/736bd3cd093cd418b4fec5b583359884f23265ff/clients/src/main/java/org/apache/kafka/clients/producer/ProducerConfig.java#L423)
- [KafkaProducer.doSend waitOnMetadata ](https://github.com/apache/kafka/blob/736bd3cd093cd418b4fec5b583359884f23265ff/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1027 )

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
@@ -215,6 +215,23 @@ sealed abstract class ProducerSettings[F[_], K, V] {
     */
   def withCredentials(credentialsStore: KafkaCredentialStore): ProducerSettings[F, K, V]
 
+  /**
+    * Controls whether [[fs2.kafka.KafkaProducer.produce]] fails immediately if any
+    * [[org.apache.kafka.clients.producer.KafkaProducer.send]] callback resolves with error.
+    *
+    * When set to `true`, the `produce` method will fail fast, returning an error as soon as any
+    * record in the [[ProducerRecords]] fails to be sent.
+    *
+    * The default value is `false`, meaning the `produce` method will not fail fast and will
+    * continue processing other records even if some callbacks fail.
+    */
+  def failFastProduce: Boolean
+
+  /**
+    * Creates a new [[ProducerSettings]] with the specified [[failFastProduce]].
+    */
+  def withFailFastProduce(failFastProduce: Boolean): ProducerSettings[F, K, V]
+
 }
 
 object ProducerSettings {
@@ -224,7 +241,8 @@ object ProducerSettings {
     override val valueSerializer: Resource[F, ValueSerializer[F, V]],
     override val customBlockingContext: Option[ExecutionContext],
     override val properties: Map[String, String],
-    override val closeTimeout: FiniteDuration
+    override val closeTimeout: FiniteDuration,
+    override val failFastProduce: Boolean = false
   ) extends ProducerSettings[F, K, V] {
 
     override def withCustomBlockingContext(ec: ExecutionContext): ProducerSettings[F, K, V] =
@@ -300,6 +318,9 @@ object ProducerSettings {
       valueSerializer: Resource[F, ValueSerializer[F, V1]]
     ): ProducerSettings[F, K1, V1] =
       copy(keySerializer = keySerializer, valueSerializer = valueSerializer)
+
+    override def withFailFastProduce(failFastProduce: Boolean): ProducerSettings[F, K, V] =
+      copy(failFastProduce = failFastProduce)
 
   }
 
--- a/modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
+++ b/modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
@@ -242,7 +242,7 @@ object ProducerSettings {
     override val customBlockingContext: Option[ExecutionContext],
     override val properties: Map[String, String],
     override val closeTimeout: FiniteDuration,
-    override val failFastProduce: Boolean = false
+    override val failFastProduce: Boolean
   ) extends ProducerSettings[F, K, V] {
 
     override def withCustomBlockingContext(ec: ExecutionContext): ProducerSettings[F, K, V] =
@@ -335,7 +335,8 @@ object ProducerSettings {
       properties = Map(
         ProducerConfig.RETRIES_CONFIG -> "0"
       ),
-      closeTimeout = 60.seconds
+      closeTimeout = 60.seconds,
+      failFastProduce = false
     )
 
   def apply[F[_], K, V](
--- a/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
@@ -7,6 +7,7 @@
 package fs2.kafka
 
 import scala.annotation.nowarn
+import scala.concurrent.Promise
 
 import cats.effect.{Async, Outcome, Resource}
 import cats.effect.syntax.all.*
@@ -140,20 +141,30 @@ object TransactionalKafkaProducer {
           sendOffsets: Option[(KafkaByteProducer, Blocking[F]) => F[Unit]]
         ): F[Chunk[(ProducerRecord[K, V], RecordMetadata)]] =
           withProducer.exclusiveAccess { (producer, blocking) =>
-            blocking(producer.beginTransaction()).bracketCase { _ =>
-              val produce = records
-                .traverse(
-                  KafkaProducer.produceRecord(keySerializer, valueSerializer, producer, blocking)
-                )
-                .flatMap(_.sequence)
+            Async[F]
+              .delay(Promise[Throwable]())
+              .flatMap { produceRecordError =>
+                blocking(producer.beginTransaction()).bracketCase { _ =>
+                  val produce = records
+                    .traverse(
+                      KafkaProducer.produceRecord(
+                        keySerializer,
+                        valueSerializer,
+                        producer,
+                        blocking,
+                        produceRecordError
+                      )
+                    )
+                    .flatMap(_.sequence)
 
-              sendOffsets.fold(produce)(f => produce.flatTap(_ => f(producer, blocking)))
-            } {
-              case (_, Outcome.Succeeded(_)) =>
-                blocking(producer.commitTransaction())
-              case (_, Outcome.Canceled() | Outcome.Errored(_)) =>
-                blocking(producer.abortTransaction())
-            }
+                  sendOffsets.fold(produce)(f => produce.flatTap(_ => f(producer, blocking)))
+                } {
+                  case (_, Outcome.Succeeded(_)) =>
+                    blocking(producer.commitTransaction())
+                  case (_, Outcome.Canceled() | Outcome.Errored(_)) =>
+                    blocking(producer.abortTransaction())
+                }
+              }
           }
 
         override def metrics: F[Map[MetricName, Metric]] =
--- a/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
@@ -141,30 +141,35 @@ object TransactionalKafkaProducer {
           sendOffsets: Option[(KafkaByteProducer, Blocking[F]) => F[Unit]]
         ): F[Chunk[(ProducerRecord[K, V], RecordMetadata)]] =
           withProducer.exclusiveAccess { (producer, blocking) =>
-            Async[F]
-              .delay(Promise[Throwable]())
-              .flatMap { produceRecordError =>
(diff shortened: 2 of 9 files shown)


aartigao:
Hey Rodrigo! Thanks for your contribution 🙌🏽 

I have some doubts about this approach, though. What if, as a user, I don't want it to fail fast? (because I don't care if some records in the batch fail?). This is valid scenario too. As an example, at $WORK we push constantly the locations of couriers coming from mobile app, and we don't want a single record to ruin the whole batch (because in a matter of seconds, a new location will be sent anyways).

But I see your point here, and it's a valid scenario too. Thing is, to me, this is a behavioral change that could potentially affect current users of the lib. And we can't introduce this kind of changes straight away. I'd be more than happy to add it as a feature under a configuration setting in `ProducerSettings` (a setting disabled by default to be compatible with all users). Something like `failFastBatch` or `failFastSend` or other naming suggestions.

Let me know what you think about it 😉

rodrigo-molina:
Hey Alan, thanks for your detailed response.

As you mentioned, I agree with the importance of not disrupting the current behavior, especially since the library supports fire-and-forget scenarios.

I'll work on making the producer fail fast feature configurable. What do you think about the name `failFastProduce`? It aligns with KafkaProducer.produce.

aartigao:
Sounds good!

rodrigo-molina:
I've updated the code and the PR description accordingly.

Please let me know your thoughts 😄

erikvanoosten:
Another relevant question: why does the kafka client not fail immediately for non-transient errors such as invalid username, password, and non-existing topics?

rodrigo-molina:
> Another relevant question: why does the kafka client not fail immediately for non-transient errors such as invalid username, password, and non-existing topics?

Thanks @erikvanoosten for chipping in.

In my opinion, non-transient errors should fail fast since they will continue to fail (and block the producer with repeated metadata request calls) until the error is resolved externally.

While this discussion is relevant to the scope of this PR, it may be interpreted as an extra feature. On the one hand, there's the idea of making the producer fail fast on any failure within the batch; on the other hand, there's the possibility of failing fast only on specific, predefined errors (assumed to be non-recoverable).

A proposal to address both is allowing users to configure which specific exceptions trigger the fail-fast behavior. This way, users can define what qualifies as a non-transient error for their specific use case and/or decide which errors to fail fast on. To simplify this for users, we could even provide a set of default error types known to be non-transient in Kafka. To avoid changing the producer's current semantics, we can disable the feature by default (no exception is failed fast by default). Things I like about this proposal:
1. The `fail fast` feature becomes highly configurable for users
2. Some defaults provide recommendations for basic use cases (like fail fast on non-transient errors) and prevent users from deep diving into some Kafka exceptions
3. Current `KafkaProducer.produce` semantics are not changed

Challenges to the proposal:
1. For _"things I like second point"_, distinguishing some errors can be tricky and non-consistent over Kafka client versions. For instance, a topic not found is a `TimeoutException` with the message "Topic $nonExistentTopic not present in metadata after $max.block.ms ms.". If we don't want to maintain this as part of the library,  we can avoid providing the defaults; the proposal still gives users enough autonomy to implement "fail fast on non-transient errors" in their context.

Finally, as an implementation detail, fail-fast errors can be provided as a partial function in the ProducerSettings. Something like:
```scala
ProducerSettings {
...
   /** 
     * Controls whether [[fs2.kafka.KafkaProducer.produce]] fails immediately if any 
     * [[org.apache.kafka.clients.producer.KafkaProducer.send]] callback resolves with error. 
     * 
     * When set to `true`, the `produce` method will fail fast, returning an error as soon as any 
     * record in the [[ProducerRecords]] fails to be sent. 
     * 
     * The default value is `false`, meaning the `produce` method will not fail fast and will 
     * continue processing other records even if some callbacks fail. 
     */ 
   def failFastProduce: Boolean 


   /** 
     * Controls which exceptions makes [[fs2.kafka.KafkaProducer.produce]] fails immediately.
     * 
     * This setting is only applied when [[failFastProduce]] is set to true.
     * 
     * The default value is to fail fast on all exceptions.
     */ 
   def failFastProduceOn: PartialFunction[Throwable, Boolean]

...
}
```

Please let me know your thoughts.

erikvanoosten:
> While this discussion is relevant to the scope of this PR, it may be interpreted as an extra feature...

Yeah I agree that this PR is a nice solution (or kludge?). We can't wait for the kafka maintainers to fix this properly (if they will at all).

modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
```diff
@@ -224,7 +241,8 @@ object ProducerSettings {
     override val valueSerializer: Resource[F, ValueSerializer[F, V]],
     override val customBlockingContext: Option[ExecutionContext],
     override val properties: Map[String, String],
-    override val closeTimeout: FiniteDuration
+    override val closeTimeout: FiniteDuration,
+    override val failFastProduce: Boolean = false
```
aartigao, line 245: Instead of defining here the default value, wouldn't it be more convenient to set it in the `create` method where the other default values are set?

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
```diff
@@ -141,11 +142,26 @@ object TransactionalKafkaProducer {
...
+                      produceRecordError
+                    )
+                  )
+                  .flatMap(_.sequence)
+
+              val produce =
+                if (settings.producerSettings.failFastProduce)
+                  Async[F]
+                    .delay(Promise[Throwable]())
+                    .flatMap(produceRecordError => produceRecords(produceRecordError.some))
```
aartigao, line 162: A `Promise` is created but never listened (if I read correctly). Is that deliberate?

erikvanoosten: The author of this PR is on holiday so I'll try to answer for him: it seems that the race that is present in modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala, was not added here. It is my guess is that this was an oversight.

rodrigo-molina: yes, the race is missing.
I'll add it.

modules/core/src/main/scala/fs2/kafka/ProducerSettings.scala
rodrigo-molina: It's addressed in commit: 5472f259d830cb6c411d8e286bc9a76c4238cd40

modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala
rodrigo-molina: good catch. It's added in commit 6442060eeb2aac91dcafc414e4d77df353f85c9f

I also added a test for it:
https://github.com/fd4s/fs2-kafka/blob/6442060eeb2aac91dcafc414e4d77df353f85c9f/modules/core/src/test/scala/fs2/kafka/TransactionalKafkaProducerSpec.scala#L479-L489

An interesting discovery is that `KafkaProducer.produce` and `TransactionalKafkaProducer.produce` behave differently. Specifically, `KafkaProducer.produce` defers the responsibility of waiting for callback resolutions to the user, as the F[F[_]] result. In contrast, `TransactionalKafkaProducer.produce` eagerly resolves the callbacks.

https://github.com/fd4s/fs2-kafka/blob/64a751e44d6a810d141e11beb8eef1ea8b532f60/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L45-L47

https://github.com/fd4s/fs2-kafka/blob/64a751e44d6a810d141e11beb8eef1ea8b532f60/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala#L41-L43

https://github.com/fd4s/fs2-kafka/blob/64a751e44d6a810d141e11beb8eef1ea8b532f60/modules/core/src/main/scala/fs2/kafka/TransactionalKafkaProducer.scala#L144-L148

https://github.com/fd4s/fs2-kafka/blob/64a751e44d6a810d141e11beb8eef1ea8b532f60/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L178-L180

note the `flatMap` and `map` difference.

modules/core/src/main/scala/fs2/kafka/producer/MkProducer.scala
```diff
@@ -17,8 +17,8 @@ import org.apache.kafka.common.serialization.ByteArraySerializer
   * the fs2-kafka `KafkaProducer`. This is needed in order to instantiate
   * [[fs2.kafka.KafkaProducer]] and [[fs2.kafka.TransactionalKafkaProducer]].<br><br>
   *
-  * By default, the instance provided by [[MkProducer.mkProducerForSync]] will be used. However this
-  * behaviour can be overridden, e.g. for testing purposes, by placing an alternative implicit
+  * By default, the instance provided by [[MkProducer.mkProducerForSync]] will be used. However,
```
rodrigo-molina, line 20: I couldn't help myself here. Please let me know if you prefer to keep this change in a different PR.

rodrigo-molina:
> September has been a crazy month for me at $work and didn't let me look much at OSS, sorry for that.
> 
> I plan to give a second round review soon. In the meantime, I'd love to hear some thoughts from other maintainers 🙏🏽

thanks @aartigao for the heads-up. No worries, I'll be looking forward to your comments.

modules/core/src/main/scala/fs2/kafka/producer/MkProducer.scala
aartigao: Hahaha, LGTM 👍🏽

aartigao (approved):
Looks good to me. 

I think I'd be helpful if other maintainers take a look too (I may have overlooked something). Pinging them just in case @vlovgr @bplommer 

@rodrigo-molina I think your PR has been waiting for quite some time now. I'd wait for 2 weeks max to see if other maintainers pop-up. If not, then do an RC for you to test. WDYT?

vlovgr (approved):

rodrigo-molina:
> Looks good to me.
> 
> I think I'd be helpful if other maintainers take a look too (I may have overlooked something). Pinging them just in case @vlovgr @bplommer
> 
> @rodrigo-molina I think your PR has been waiting for quite some time now. I'd wait for 2 weeks max to see if other maintainers pop-up. If not, then do an RC for you to test. WDYT?

Thanks @aartigao for the review and for helping us move this forward.

We already have another approval 🙌 

Ping me when you create the release candidate, and we'll run it in production.

aartigao:
> Ping me when you create the release candidate, and we'll run it in production.

Just released 3.6.0-RC1.

rodrigo-molina:
> > Ping me when you create the release candidate, and we'll run it in production.
> 
> Just released 3.6.0-RC1.

Awesome!
I'll update it in our projects tomorrow and let you know how it goes.

rodrigo-molina:
> > Ping me when you create the release candidate, and we'll run it in production.
> 
> Just released 3.6.0-RC1.

it's already in production and working nicely. Now, faulty integrations fail in the orther of milliseconds instead of half of a minute when missing grants (a 30-second timeout was hardcoded before having this)!

Thanks for the support!

PR #1367, opened 2024-12-18
chore: make state transitions uncancelable

biochimia:
Ensures updates to `KafkaConsumerActor`'s internal state are applied together with the side-effects that support those updates by executing both together inside an uncancelable block. These changes ensure that changes made to the consumer are reflected in the internal state, and changes to the internal state are made visible to clients.

In the simplest cases, a sequence of `Ref.update`/`modify` and `Async.flatMap` was refactored to use `Ref.flatModify`, which offers the desired guarantees. In other cases, the relevant actions were explicitly wrapped in an `Async.uncancelable` block.

Where relevant, logging of state changes was treated as part of the uncancelable block. This is to avoid silent state changes, which may be hard to track down while debugging. Invocation of user-supplied callbacks on changes to assigned partitions is left out of these blocks as failures in those do not affect the internal state of the consumer.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -185,7 +185,7 @@ object KafkaConsumer {
                     def storeFetch: F[Unit] =
                       actor
                         .ref
-                        .modify { state =>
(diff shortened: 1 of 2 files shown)


aartigao (approved):
Nice! 👍🏽

PR #1368, opened 2024-12-18
chore: tweaks to handling of pending commits

biochimia:
This is a combination of two related changes to the handling of pending commits during rebalance operations in `KafkaConsumerActor`.

### Single definition of pending commit effect

Whether immediately committing offsets, or queueing the request in `State.pendingCommits`, the necessary effect is built at once.

This avoids the need to repeat similar logic in disparate call sites, and makes visible in a single place how the request will be processed and logged.

### Ensure pending commits order

Commits would be added to the chain of pending commits based only on the rebalancing state. Given that the rebalancing state is updated (via `ConsumerRebalanceListener.onPartitionsAssigned`) separate from when pending commits are processed (after `poll`), it could happen that commits emitted later would be processed before earlier pending ones.

This updates the condition for queueing commits to take into account the prior existence of pending commits.

In addition, the condition for processing pending commits in `poll` is also updated to disregard whether a rebalance operation was ongoing at the start of the poll. Instead, the existence of pending commits along with a non-`rebalancing` state are a sufficient trigger.

This ensures that rebalance operations that might conclude within a single consumer poll do not leave behind any pending commits.

At the moment, these possibilities are theoretical as commit operations are serialized via `KafkaConsumerActor`'s request queue, and don't happen concurrently to polls. That said, the cost of the fixes is trivial and being explicit about the conditions may prevent future bugs, if the surrounding context changes.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/internal/LogEntry.scala
+++ b/modules/core/src/main/scala/fs2/kafka/internal/LogEntry.scala
@@ -145,7 +145,7 @@ private[kafka] object LogEntry {
 
     override def level: LogLevel = Debug
 
-    override def message: String = s"Committed pending commits [$pendingCommit]."
+    override def message: String = s"Committed pending commit [$pendingCommit]."
 
   }
 
--- a/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
+++ b/modules/core/src/main/scala/fs2/kafka/internal/KafkaConsumerActor.scala
@@ -93,7 +93,7 @@ final private[kafka] class KafkaConsumerActor[F[_], K, V](
   private[this] def commit(request: Request.Commit[F]): F[Unit] =
     ref.flatModify { state =>
       val commitF = commitAsync(request.offsets, request.callback)
-      if (state.rebalancing) {
+      if (state.rebalancing || state.pendingCommits.nonEmpty) {
         val newState =
           state.withPendingCommit(commitF >> logging.log(CommittedPendingCommit(request)))
         (newState, logging.log(StoredPendingCommit(request, newState)))


modules/core/src/main/scala/fs2/kafka/internal/LogEntry.scala
```diff
@@ -211,15 +211,11 @@ private[kafka] object LogEntry {
...
-    pendingCommits: Chain[Request.Commit[F]],
-    state: State[F, ?, ?]
-  ) extends LogEntry {
+  final case class CommittedPendingCommit[F[_]](pendingCommit: Request.Commit[F]) extends LogEntry {
 
     override def level: LogLevel = Debug
 
-    override def message: String =
-      s"Committed pending commits [$pendingCommits]. Current state [$state]."
+    override def message: String = s"Committed pending commit [$pendingCommit]."
```
aartigao, line 218: Why did you exclude the logging of the state?

aartigao (commented):
I think the PR has conflicts from another PR of yours. Can you fix them, please? 🙏🏽

modules/core/src/main/scala/fs2/kafka/internal/LogEntry.scala
biochimia: In part, the reason for dropping logging of the state here is that the updated state is not available at the point the commit effect is prepared in the `KafkaConsumerActor.commit`.

Additionally, with this change, and the way the commits are processed, the log message is now made per pending commit, and not just once for the full chain. (There will be one `CommittedPendingCommit` log entry per `StoredPendingCommit` entry.) At this granularity, it might be noisy to repeat the updated state.

Now, none of this actually requires that the updated state be dropped from the log message, even if it was convenient to do so.

Going over where pending commits are handled, in `KafkaConsumerActor.handlePoll`, the updated state should be logged prior to the handling of pending commits, so we shouldn't lose detail in the logs. Considering this versus the added complexity  to kep the log messages as before, it seemed like a good trade-off to drop the detail.

biochimia:
I have rebased my changes on top of the updated branch.

aartigao:
More conflicts after merging your changes.

biochimia:
Thank you for taking the time to review and merge this.

The other PR included and obsoleted some of these changes, so this one became really small.

I developed all the patches together, and split them in the end, as I expected the `FetchRequest` changes to warrant a deeper dive and wanted to keep noise to the minimum on that one. Conflicts were an unavoidable consequence.

PR #1369, opened 2024-12-18
feat: drop use of FetchRequests

biochimia:
Record streams are streamlined by dropping the `FetchRequest` middleman. Instead, queues of records are lazily created by `KafkaConsumerActor`, one for each assigned partition.

Enqueueing records uses `Queue.tryOffer`, to avoid blocking the polling fiber. `KafkaConsumerActor` maintains a separate `spillover` map to hold records when queues are full. This acts as back-pressure and pauses polling for "slow" partitions.

`KafkaConsumerActor.poll` takes a more active role in managing assigned partitions. Namely, it signals the revocation of any partitions that are not assigned, and drops those from the internal state.

While the `ConsumerRebalanceListener` interface is meant to handle this state when the consumer faces rebalance operations, explicitly handling assignments in `poll` caters to manually assigned partitions. In addition, it ensures `KafkaConsumerActor`'s internal state stays consistent in the face of race conditions between rebalance operations and the setup of record streams in `KafkaConsumer`.

The newly introduced `partitionState` (maintaining per-partition queues and spillover records) bears some resemblance to the former `fetches` and `records` fields, but differs in some important ways:

- registration of fetch requests involved a query of the current consumer assignment, which forced synchronization with the polling thread via the use of `withConsumer.blocking`.
- fetch requests would follow the lifcycle of record chunks, and needed to be continuously re-added to the state, before new data would be polled from a partition. Now, the current assignment forms the basis to fetch data from a partition, with spillover records acting as a back-pressure signal;
- `records` acted as a holding area, but mainly supported a race condition between a new assignment, the start of a new stream, and the subsequent registration of fetch requests. `records` was not generally used as a spillover area for incoming data.

In the face of multiple streams hooked up to a single consumer, an inherent race in the old registration of fetch requests meant that each chunk of records could be added to all or only a subset of the listening streams. With the new approach, multiple streams will forcefully compete to take elements from the queue, ensuring each chunk of records goes to only one stream. (While I do not expect that such use of multiple streams was a feature, the potential behavior change is noted.)

An internal `StreamId` was previously used to match `FetchRequest`s to the owning stream. This is no longer used and the ID is dropped.

`KafkaConsumer` previously kept track of its partition assignment, along with a `Deferred` instances to interrupt partition streams. This is now handled by `KafkaConsumerActor`, which relies solely on the underlying consumer to keep track of its current assignment.

Overall, this change should reduce some overhead and latency between the polling thread and the record streams. It does this by removing layers through which records must pass, as well as reducing synchronization between polling and consuming fibers.

### DIFF
--- a/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
+++ b/modules/core/src/main/scala/fs2/kafka/KafkaConsumer.scala
@@ -27,7 +27,6 @@ import fs2.kafka.internal.*
 import fs2.kafka.internal.converters.collection.*
 import fs2.kafka.internal.syntax.*
 import fs2.kafka.internal.KafkaConsumerActor.*
-import fs2.kafka.internal.LogEntry.{RevokedPreviousFetch, StoredFetch}
 
 import org.apache.kafka.clients.consumer.{OffsetAndMetadata, OffsetAndTimestamp}
 import org.apache.kafka.common.{Metric, MetricName, PartitionInfo, TopicPartition}
@@ -127,10 +126,8 @@ object KafkaConsumer {
 
   private def createKafkaConsumer[F[_], K, V](
     requests: QueueSink[F, Request[F, K, V]],
-    settings: ConsumerSettings[F, K, V],
     actor: KafkaConsumerActor[F, K, V],
     fiber: Fiber[F, Throwable, Unit],
-    streamIdRef: Ref[F, StreamId],
     id: Int,
     withConsumer: WithConsumer[F],
     stopConsumingDeferred: Deferred[F, Unit]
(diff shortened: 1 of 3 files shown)


aartigao:
I really like the proposal of this PR. Currently there are too many indirection layers for something that can be simpler, as this demonstrates. Fix the conflicts and the PR is good to go.

I have to admit that I'm slightly concerned about so core changes so I plan to do a RC with them. This way we can have some artifacts to test in our companies staging envs, just in case. Even if the tests are green, Kafka stuff with the workflows like rebalancing have sneaky edge cases that usually pop up in real life scenarios.

biochimia:
I have rebased my changes and resolved the conflicts.

biochimia:
Thank you @aartigao.

When you cut a RC we can take it for a spin in Adevinta where we have an application with thousands of consumers.

aartigao:
Just published 3.7.0-RC1.

I'll try it too in our company's Staging Environment with one of our core projects 🙃 It's not thousands of consumers but still... 😛 

Thank you for your patience both you and Rodrigo 🙏🏽  I'm having crazy months since my company is sinking and a lot of people either left o was fired and the rest have to assume their job 🥲 And having free time outside working hours is almost impossible with a baby 😂

biochimia:
Thanks again, @aartigao.

No worries from our side. We fully understand this is volunteer work. Unfortunately, neither of us can commit to more than the occasional comment and patch either. These are equally interesting times at Adevinta.

Be sure to take care of yourself and your family first!

biochimia:
Sorry for the radio silence. Last week, I didn't have a chance to pick this up. The last couple of days I deployed and ultimately rolled back the release candidate a couple of times on our side.

The reason for rolling back is that we're seeing some errors when fetching Avro schemas from the schema registry, that we don't see with the previous release. The errors and the affected code don't seem to bear a direct relation to changes in the release or to `fs2-kafka` itself, as we handle deserialisation separately, and outside `fs2.kafka.KafkaConsumer`. That said, the errors do somehow show up with the updated package.

I spent some time updating and deploying dependencies separately from the 3.7.0 release candidate, without luck. So far only updating `fs2-kafka` triggers the errors. I need to double check if I missed some dependency.

Drilling into the errors hasn't enlightened me, so far, and is ongoing.

The errors we saw affect a small proportion of our consumers and volume of data. Most consumers seemed to be processing data just fine during the short time we had the release candidate running. That said we will need more time to understand what is going on.

aartigao:
That's intriguing indeed. We also use AVRO messages with the official Confluent serdes:

- `io.confluent.kafka.serializers.KafkaAvroDeserializer`
- `io.confluent.kafka.serializers.KafkaAvroSerializer`

and I haven't seen any issues so far. But it's true that I've deployed the RC on a service that consumes 6 o 7 schemas from Schema Registry (might be different from yours).

Maybe you can make a dependency diff to see if the Confluent deps bumped or have conflicts with the RC? 🤔

biochimia:
I found a bug with my changes in the handling of spillover records. I'll submit a new PR for the issue.

biochimia:
New PR submitted: #1398.

The issue I was seeing with schemas affected consumers moving Avro data between different Kafka clusters and also translating schema IDs between different Schema Registry instances. The schema IDs are being updated in-place in raw `Array[Byte]` records.

The bug meant that spillover records could be presented twice, and because the records are updated in place the second time around the schema IDs were already translated to the target schema registry and didn't match schemas in the source schema registry...

😅

biochimia:
For other use cases that don't update records in-place, the bug would show up as duplicate records.

Issue #1391, opened 2025-02-27
Promise[Throwable] fullfilled multiple times in KafkaProducer


ex-ratt:
I made some tests regarding rebalancing, where I used `KafkaConsumer.partitionsMapStream` with `Stream.switchMap` to interrupt the processing of streams whenever a rebalance happens (to prevent processing buffered records that may have been assigned to a different consumer). In one test, the interruption just happened to be in the moment that some records were being (transactionally) produced. For each producer, there were n records to produce and (n - 1) such exceptions:

```
11:21:04.522 ERROR: Error executing user-provided callback on message for topic-partition 'test-pstate-output-1'
java.lang.IllegalStateException: Promise already completed.
    at scala.concurrent.Promise.complete(Promise.scala:57)
    at scala.concurrent.Promise.complete$(Promise.scala:56)
    at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:104)
    at scala.concurrent.Promise.failure(Promise.scala:109)
    at scala.concurrent.Promise.failure$(Promise.scala:109)
    at scala.concurrent.impl.Promise$DefaultPromise.failure(Promise.scala:104)
    at fs2.kafka.KafkaProducer$.$anonfun$produceRecord$7(KafkaProducer.scala:221)
    at scala.Option.foreach(Option.scala:437)
    at fs2.kafka.KafkaProducer$.$anonfun$produceRecord$6(KafkaProducer.scala:221)
    at org.apache.kafka.clients.producer.KafkaProducer$AppendCallbacks.onCompletion(KafkaProducer.java:1567)
    at org.apache.kafka.clients.producer.internals.ProducerBatch.completeFutureAndFireCallbacks(ProducerBatch.java:311)
    at org.apache.kafka.clients.producer.internals.ProducerBatch.abort(ProducerBatch.java:199)
    at org.apache.kafka.clients.producer.internals.RecordAccumulator.abortUndrainedBatches(RecordAccumulator.java:1161)
    at org.apache.kafka.clients.producer.internals.Sender.maybeSendAndPollTransactionalRequest(Sender.java:473)
    at org.apache.kafka.clients.producer.internals.Sender.runOnce(Sender.java:336)
    at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:250)
    at java.lang.Thread.run(Thread.java:1575)
```

[...]

PR #1393, opened 2025-03-09
Fix exeption of multiple promise completions (#1391)

ex-ratt:
Using `tryFailure` instead of `failure` will not produce an `IllegalStateException` when the sending of multiple records is interrupted. The same promise is used for all records and all of them will complete that same promise, leading to the exception without this change.

This fixes issue #1391.


aartigao (approved):
Thanks for spotting the issue and provide a fix!

Issue #1200, opened 2023-05-11
Multiple consumers read from the same partition after rebalance


squadgazzz:
With fs2-kafka `3.0.1` and also on `2.x.x`, we faced with the issue where multiple consumers that share the same group id start reading from the same partition after rebalance.
The flow looks as follows:
- Consumer A is reading from partition 1. Current offset is 1453;
- A new consumer B connects and is assigned to partition 1;
- Logs for the consumer A: 
```
Updating assignment with
Assigned partitions: [mytopic.v1-2]
Current owned partitions: [mytopic.v1-2, mytopic.v1-1]
Added partitions (assigned - owned): [] Revoked partitions (owned - assigned): [mytopic.v1-1]
Revoke previously assigned partitions mytopic.v1-1
Need to revoke partitions [mytopic.v1-1] and re-join the group
Notifying assignor about the new Assignment(partitions=[mytopic.v1-2])
Adding newly assigned partitions:
(Re-)joining group
Failing OffsetCommit request since the consumer is not part of an active grou
Successfully joined group with generation Generation{generationId=289, memberId=***protocol='cooperative-sticky'}
Successfully synced group in generation Generation{generationId=289, memberId=***protocol='cooperative-sticky'}
Updating assignment with
Assigned partitions: [mytopic.v1-2]   
Current owned partitions: [mytopic.v1-2]
Added partitions (assigned - owned): [] Revoked partitions (owned - assigned): []
Notifying assignor about the new Assignment(partitions=[mytopic.v1-2])
Adding newly assigned partitions
```
- Consumer B starts reading from offset 1455, but consumer A continues reading from the same partition for some reason. It fails to process the event multiple times and moves to the next one, and so on. 
- Sometime after, the offset for consumer A is around 1600 and for consumer B is 2000. Another rebalance happens and consumer A gets the ownership again. Since the offset was 1600 it successfully commits that offset and starts reading the event up to 2000 that were already processed by consumer B.

We have already checked the broker and found no issue. Especially the logs confirm that the broker issues the right events. There might be something with the `ConsumerRebalanceListener` implementation in the library.

squadgazzz:
Might be related to the https://github.com/fd4s/fs2-kafka/issues/127

wookievx:
We had very similar issue at my workplace that old consumer kept consuming after being revoked. After investigating fs2-kafka implementation and comparing it to the changes that were done in the zio-kafka I am not sure this is a root cause. 
[...]

PR #1399, opened 2025-04-02
feat: Implemented RebalanceRevokeMode

wookievx:
This is potential solution for https://github.com/fd4s/fs2-kafka/issues/1200. Allowing for opting-in for graceful revoke handling (waiting for all the streams to finish, which should imply all the commits as well)


simonpetty:
Hope you don't mind a quick question here. Why re-use the session timeout, instead of a separate value? (We have quite a high session timeout, as we're using cooperative partition assigners).

PR #1405, opened 2025-05-12
removed `retries` value override

Awethon:
This was introduced due to the default behavior of Kafka prior to version 3.0. However, it now causes more harm than good by preventing fail-safe behavior during rebalancing, as `NotLeaderOrFollowerException` is not retried.

Since Kafka 3.0, the default is `enable.idempotence=true`, which ensures safe behavior according to both the SoftwareMill blog (mentioned in producers.md) and the [Kafka docs](https://kafka.apache.org/documentation/#producerconfigs_retries).

> Allowing retries while setting enable.idempotence to false and max.in.flight.requests.per.connection to greater than 1 will potentially change the ordering of records because if two batches are sent to a single partition, and the first fails and is retried but the second succeeds, then the records in the second batch may appear first.

### DIFF
--- a/docs/src/main/mdoc/producers.md
+++ b/docs/src/main/mdoc/producers.md
@@ -127,10 +127,6 @@ ProducerSettings(
 
 ### Default Settings
 
-The following Java Kafka producer properties are overridden by default.
-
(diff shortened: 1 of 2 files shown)


Awethon:
After reading more kafka code, I think this override is even more harmful than I previously thought as it disables idempotence and makes kafka producer less reliable:
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/ProducerConfig.java#L580-L593

Awethon:
@vlovgr maybe you can take a look if possible? 🙏 Can't find any other active maintainer.

sideeffffect:
Thank you so much @vlovgr can you also make a new release with this fix?

vlovgr:
@sideeffffect I've released [v3.8.0](https://github.com/fd4s/fs2-kafka/releases/tag/v3.8.0) with these changes.

PR #1486, opened 2026-05-09
Add otel4s-trace module

iRevive:
Hey everyone! I'm trying to bring [otel4s](https://github.com/typelevel/otel4s) tracing support in this PR.

> [!NOTE]
> The PR was implemented with the help of AI. However, the final version was thoroughly reviewed by a human (me).

### 1. Switched the build to [sbt-projectmatrix](https://github.com/sbt/sbt-projectmatrix)

otel4s is available only for Scala 2.13 and 3, while fs2-kafka supports 2.12, 2.13, and 3. 
Implications: now there is **1** CI job that runs tests for **all** targets.

### 2. The `otel4s-trace` module

The module covers tracing operations only. Perhaps we might want to implement `-metrics` too at some point.
Currently, only `send`, `create`, `receive`, and `process` operations are covered.

### 1. Is exposing producer/consumer metadata in the core module acceptable?

The tracing module needs producer/consumer metadata like `client.id` and `group.id`. Right now, that requires two small package-private methods in the core. 

Is it fine, or should we drop this logic entirely and have users configure it via const attributes?

### 2. Should we allow configuration batch spans?

For multi-record sends, the implementation emits **per-record** `create` spans to cleanly preserve creation context. That is semantically useful, but it can be expensive in high-throughput cases. 

Should we have a config option to control the shape of said spans?
    
### 3. Should the traced API stay explicit and separate from `KafkaProducer` / `KafkaConsumer`?

The current design keeps tracing explicit: users bind a `KafkaTracer` to a producer/consumer and call traced operations directly. That avoids hidden behavior and makes receive / process boundaries clear, but it is less drop-in than mirroring the full producer/consumer API. 

`TracedKafkaProducer` can extend `KafkaProducer`, but I don't think `TracedKafkaConsumer` can extend `KafkaConsumer` and implement instrumentation logic for **all** operations.

### DIFF
--- a/modules/otel4s-trace/src/main/scala/fs2/kafka/otel4s/trace/TracedKafkaConsumer.scala
+++ b/modules/otel4s-trace/src/main/scala/fs2/kafka/otel4s/trace/TracedKafkaConsumer.scala
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2018 OVO Energy Limited
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package fs2.kafka.otel4s.trace
+
+import scala.util.chaining._
+
+import cats.effect.Concurrent
+import cats.syntax.all._
+import fs2.{Chunk, Stream}
+import fs2.kafka.{CommittableConsumerRecord, ConsumerRecord, KafkaConsumer}
+import fs2.kafka.otel4s.trace.instances._
+import fs2.kafka.otel4s.trace.internal.Semconv
+
+import org.typelevel.otel4s.trace.{SpanContext, SpanKind, Tracer}
+import org.typelevel.otel4s.Attributes
+
+/**
+  * A consumer-bound tracing handle for `fs2-kafka`.
+  *
+  * Unlike producer tracing, consumer tracing cannot be a fully transparent drop-in replacement,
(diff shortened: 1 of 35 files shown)


modules/otel4s-trace/src/main/scala/fs2/kafka/otel4s/trace/TracedKafkaConsumer.scala
```diff
@@ -0,0 +1,194 @@
...
+
+object TracedKafkaConsumer {
+
+  final private[otel4s] class Impl[F[_]: Concurrent: Tracer, K: KafkaMessageKey, V](
+    override val underlying: KafkaConsumer[F, K, V],
+    config: KafkaTracer.Config
+  ) extends TracedKafkaConsumer[F, K, V] {
+
+    override def records: Stream[F, CommittableConsumerRecord[F, K, V]] =
+      underlying.records
```
iRevive, line 101: Since there is no instrumentation, perhaps we can drop the method?

vlovgr:
Following [discussions](https://discord.com/channels/632277896739946517/909159046559465545/1502658929203613716) on Discord, I've raised https://github.com/typelevel/fs2-kafka/pull/1487 to enable this to exist as a separate library.
Once there's a release available, feel free to update the documentation with a link to the library.

